PR #2, opened 2020-06-08
Adding parametric shapes

billingsley-john:


billingsley-john:
This PR repeated some commits from PR #1

PR #25, opened 2020-06-24
Update parametric shapes

billingsley-john:
PR updating parametric shapes. removes unnecessary re-defining of solid getters and setters in the parametric shape classes when they are already inherited from user classes by the super.
Also some minor updates to docstrings and property inheritance.


Shimwell:
Thanks @billingsley-john nice code reductions here

PR #42, opened 2020-07-03
updated import statements

Shimwell:
This restructures the modules so that importing is done in a more pythonic manner.

Currently we import like this

from paramak import paramaetric_shapes.CenterColumnShieldCylinder()

and use like this

CenterColumnShieldCylinder()

with this PR we can import like this

import paramak as pm

and use like this

pm.
CenterColumnShieldCylinder()


Shimwell:
I have also updated the examples so that they use this imporved method

and I might have got carried away and added a neutronics simulation to the local tests and a neutronics simulation to the examples, perhaps this was out of scope for the PR. Sorry about that

PR #50, opened 2020-07-09
Moving examples around

billingsley-john:
PR which organizes example scripts into 4 sub-directories, edits the scripts so that they are not executed by sphinx and adds docstrings to the scripts


billingsley-john:
Also adds documentation for the example scripts

Shimwell:
Looks fine to me , there are a few cases where the file rename has not been noticed but it has been deleted and recreated. But nothing major. Looking forward to seeing how read the docs builds it

PR #58, opened 2020-07-11
New parametric blanket : D shaped

RemiTheWarrior:
This PR includes a new D shaped parametric blanket component as shown below.

```python
from paramak import BlanketConstantThicknessD

shape = BlanketConstantThicknessD(height=10, thickness=1.5, minor_radius=1, rotation_angle=180)
shape.solid
```

This allows to create blankets components for non-spherical tokamaks.

image

### DIFF
--- a/paramak/parametric_components/blanket_constant_thickness_D.py
+++ b/paramak/parametric_components/blanket_constant_thickness_D.py
@@ -8,10 +8,32 @@ from paramak import RotateMixedShape
 
 class BlanketConstantThicknessD(RotateMixedShape):
     """
-
+    :param height:  the height of the inner straight part of the blanket (cm)
+    :type height: float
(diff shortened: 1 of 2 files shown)


RemiTheWarrior:
Just noticed i had to make room for divertor. Closing this for now.

Shimwell:
Hi @RemiTheWarrior, missing the divertor is not so bad as this can be cut out afterwards.. I was wondering about the minor radius and how this controls the shape. I shall keep tinkering and try to understand it a bit more.

This sort of combined inboard and outboard blanket would be a great addition. Currently it has to be made in two stages which is not ideal.

RemiTheWarrior:
@Shimwell 
I'll add something a bit better I think.
In this PR, `minor_radius` was the red arrow (see below). Maybe not the right variable name for it ...
image

PR #72, opened 2020-07-15
Update for new plasma

Shimwell:
Just a small update which gets one of the examples working again (make_plasmas.py) and starts fixing the parametric reactors.


Shimwell:
this is not quite read yet, a few more things to tidy up

PR #74, opened 2020-07-15
updated pypi action

billingsley-john:
New PR to test github actions for updating python package on pypi

### DIFF
--- a/.github/workflows/python-publish.yml
+++ b/.github/workflows/python-publish.yml
@@ -24,10 +24,12 @@ jobs:
       run: |
         python -m pip install --upgrade pip
         pip install setuptools wheel twine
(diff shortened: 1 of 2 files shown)


billingsley-john:
This implementation of bump means that the version is bumped from the version specified in the setup.py file. This will be branch-dependent. As a result, some PR's may fail to trigger a new pip build depending on when the branch was created relative to other branches/merges.

billingsley-john:
As merges from develop into main will be quite rare, it is not unreasonable to remember to manually bump the version before a merge is made. This is currently the preferred solution.

billingsley-john:
This was a test PR to establish a viable workflow for pypi updating.

PR #79, opened 2020-07-15
updated pypi publish criteria

billingsley-john:
Updates yml file so pypi is updated when a new release is created


billingsley-john:
This commit causes a breakage somewhere which prevents automatic merging. Abandoning merge

PR #83, opened 2020-07-16
Building block of FE simulations: look-up table for exporting physical groups

RemiTheWarrior:
**This PR aims to introduce ways for exporting information (readable by humans) on physical groups in step files.**

### Current state:
If a user wants to use the generated step file of say a `BlanketConstantThicknessFP()` in a FE program, he/she needs to guess (or trial/error):
- the number of physical groups (physical surfaces and volumes)
- which id (already generated by CADquery) corresponds to which physical group

### This PR:
Classes inheriting from `Shape()` can now have an attribute called `look_up_table` which contains a list of dicts with physical groups infos: 
```python
look_up_table = [
    {"dim": 3, "id": 1, "name": "inside_volume"},
    {"dim": 2, "id": 1, "name": "top_surface"},
    {"dim": 2, "id": 2, "name": "inner_surface"},
    {"dim": 2, "id": 2, "name": "outer_surface"},
]
```

Above, the relations between "id" and "name" have to be determined in `BlanketConstantThicknessFP()` before hand.

This look-up table is exported in a JSON file when the STEP file is generated.
Now a user willing to use the STEP file in a FE program can rely on this table to apply boundary conditions and material properties without having to spend time in trial/error.

### Example usage with GMSH
```python
import gmsh
import json
gmsh.initialize()
gmsh.option.setNumber("General.Terminal", 1)  # for message printing


def generate_mesh(stepfile, lookup_table, meshfile="mesh", directory="."):

    # gmsh.merge(stepfile)
    gmsh.model.occ.importShapes(stepfile)
    gmsh.model.occ.synchronize()
    for group in lookup_table:
        dim = group["dim"]
        id_ = group["id"]
        gmsh.model.addPhysicalGroup(dim, [id_])
        gmsh.model.setPhysicalName(dim, id_, group["name"])

    gmsh.model.mesh.generate(3)

    gmsh.write("{}/{}.msh".format(directory, meshfile))
    gmsh.finalize()
    return


# LOAD LOOK-UP TABLE
with open('step_file_look_up_table.json') as f:
  look_up_table = json.load(f)

# GENERATE MESH
generate_mesh("step_file.stp", lookup_table)

```

### ToDo:
- Create these look-up tables for parametric components (Defaulting to `None` so that it doesn't break any test)
- Discuss this with a super developping team

### DIFF
--- a/paramak/reactor.py
+++ b/paramak/reactor.py
@@ -164,7 +164,7 @@ class Reactor():
                                  Shapes before using the export_stp method"
                 )
             filenames.append(str(Path(output_folder) / Path(entry.stp_filename)))
-            entry.export_stp(Path(output_folder) / Path(entry.stp_filename), export_look_up_table=export_look_up_tables)
+            entry.export_stp(Path(output_folder) / Path(entry.stp_filename))
 
         # creates a graveyard (bounding shell volume) which is needed for nuetronics simulations
         self.make_graveyard()
--- a/paramak/reactor.py
+++ b/paramak/reactor.py
@@ -146,7 +146,7 @@ class Reactor():
 
         return filename
 
-    def export_stp(self, output_folder="", export_look_up_tables=False):
+    def export_stp(self, output_folder=""):
         """Writes stp files (CAD geometry) for each Shape object in the reactor
 
         :param output_folder: the folder for saving the stp files to
--- a/paramak/reactor.py
+++ b/paramak/reactor.py
@@ -177,6 +177,32 @@ class Reactor():
 
         return filenames
 
+    def export_physical_groups(self, output_folder=""):
+        """Exports several JSON files containing a look up table
+        which is useful for identifying faces and volumes. The
+        output file names are generated from .stp_filename properties.
+
+        Args:
+            output_folder (str, optional): directory of outputfiles.
+                Defaults to "".
+
+        Raises:
+            ValueError: if one .stp_filename property is set to None
+
+        Returns:
+            list: list of output file names
+        """
+        filenames = []
+        for entry in self.shapes_and_components:
+            if entry.stp_filename is None:
+                raise ValueError(
+                    "set .stp_filename property for \
+                                 Shapes before using the export_stp method"
+                )
+            filenames.append(str(Path(output_folder) / Path(entry.stp_filename)))
+            entry.export_physical_groups(Path(output_folder) / Path(entry.stp_filename))
+        return filenames
+
     def export_graveyard(self, filename="Graveyard.stp"):
         """Writes a stp file (CAD geometry) for the reactor graveyard.
            Thich is needed for DAGMC simulations
--- a/paramak/reactor.py
+++ b/paramak/reactor.py
@@ -146,7 +146,7 @@ class Reactor():
 
         return filename
 
-    def export_stp(self, output_folder=""):
+    def export_stp(self, output_folder="", export_look_up_tables=False):
         """Writes stp files (CAD geometry) for each Shape object in the reactor
 
         :param output_folder: the folder for saving the stp files to
@@ -164,7 +164,7 @@ class Reactor():
                                  Shapes before using the export_stp method"
                 )
             filenames.append(str(Path(output_folder) / Path(entry.stp_filename)))
-            entry.export_stp(Path(output_folder) / Path(entry.stp_filename))
+            entry.export_stp(Path(output_folder) / Path(entry.stp_filename), export_look_up_table=export_look_up_tables)
 
         # creates a graveyard (bounding shell volume) which is needed for nuetronics simulations
         self.make_graveyard()
--- a/paramak/shape.py
+++ b/paramak/shape.py
@@ -314,7 +314,7 @@ class Shape:
 
         return str(Pfilename)
 
-    def export_stp(self, filename=None, export_look_up_table=False):
+    def export_stp(self, filename=None:
         """Exports an stp file for the Shape.solid.
         If the provided filename doesn't end with
         .stp or .step then .stp will be added. If a
--- a/paramak/shape.py
+++ b/paramak/shape.py
@@ -71,7 +71,7 @@ class Shape:
         self.render_mesh = None
         # self.volume = None
 
-        self.look_up_table = None
+        self.physical_groups = None
 
     @property
     def workplane(self):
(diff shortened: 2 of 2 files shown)


Shimwell:
I understand now that I see the code. We have something similar for neutronics. Users can export_neutronics_description which makes a JSON file with a few properties.

Just wondering if this needs to be exported with every step file export. Perhaps it could have its own export

Shimwell:
Or perhaps the export_stp could have a flag to allow such groups to be exported as well. This could default to false

RemiTheWarrior:
@Shimwell you're right, it doesn't have to be exported every time. In the current state, if no look_up table is specified it won't be exported. But I will add a flag to export_stp.

RemiTheWarrior:
@Shimwell should json be added to requirements.txt ?

Shimwell:
> @Shimwell should json be added to requirements.txt ?

JSON is in the stardard python library so we don't need to add it to the requirements

Shimwell:
Have you seen reactor.py also has an export_stp which calls the shape.py export_stp.

Is it worth adding a flag there as well

RemiTheWarrior:
@Shimwell it'd be great if the CI workflow was run at each PR
@billingsley-john

Shimwell:
> @Shimwell it'd be great if the CI workflow was run at each PR
> @billingsley-john

Looks like there is a pypi related problem, I know John is looking into this next week

paramak/shape.py
```diff
@@ -69,6 +72,8 @@ def __init__(
         self.render_mesh = None
         # self.volume = None
 
+        self.look_up_table = look_up_table
```
Shimwell:
```suggestion
        self.look_up_table = None
```

paramak/shape.py
```diff
@@ -50,6 +52,7 @@ def __init__(
         stp_filename=None,
         azimuth_placement_angle=0,
         workplane="XZ",
+        look_up_table=None,
```
Shimwell:
```suggestion
```

paramak/shape.py
```diff
@@ -331,13 +335,25 @@ def export_stp(self, filename=None):
...
+            if self.look_up_table is not None:
+                if filename.endswith('.stp'):
+                    prefix = filename[:-4]
+                json_filename = '{}_lookup_table.json'.format(prefix)
+                with open(json_filename, 'w+') as outfile:
+                    json.dump(self.look_up_table, outfile)
+                print("Saved look-up table as {}".format(json_filename))
+            else:
+                print("Look-up table not found for shape {}".format(self.name))
+
```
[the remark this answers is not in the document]
RemiTheWarrior: This should not just be removed but replaced by a function call ?

paramak/shape.py
```diff
@@ -331,13 +335,25 @@ def export_stp(self, filename=None):
...
         return str(Pfilename)
```
Shimwell, line 358:
```suggestion
    def export_look_up_table(self, filename):
        """Exports a JSON file containing a look up table
        which is useful for identifying faces. If provided
        filename doesn't end with .json then .json will be added.

        :param filename: the filename to save the json look up table
        :type filename: str

        :param filename: the filename of the json file
        :type filename: str
        """

        Pfilename = Path(filename)

        if Pfilename.suffix != ".json":
            Pfilename = Pfilename.with_suffix(".json")

        Pfilename.parents[0].mkdir(parents=True, exist_ok=True)

        with open(filename, "w") as outfile:
            json.dump(self.look_up_table(), outfile, indent=4)

        print("saved geometry description to ", Pfilename)

        return filename
```

Shimwell:
Sorry I'm slow on this, there is just one thing that is bugging me

Including the export_look_up_table within the export_stp file has the problem that the user can't really specify the filename of the export_look_up_table json file. It is currently based on the stp file name. I much prefer the separate export_look_up_table function as it keeps distinct operations separate. I don't think we bundle exports together for anything else and we have quite a few exports. json files, png, svg, stp, stl.

paramak/reactor.py
```diff
@@ -146,7 +146,7 @@ def export_neutronics_description(self, filename="manifest.json"):
 
         return filename
 
-    def export_stp(self, output_folder=""):
+    def export_stp(self, output_folder="", export_look_up_tables=False):
```
Shimwell, line 149:
```suggestion
    def export_stp(self, output_folder=""):
```

paramak/reactor.py
```diff
@@ -164,7 +164,7 @@ def export_stp(self, output_folder=""):
                                  Shapes before using the export_stp method"
                 )
             filenames.append(str(Path(output_folder) / Path(entry.stp_filename)))
-            entry.export_stp(Path(output_folder) / Path(entry.stp_filename))
+            entry.export_stp(Path(output_folder) / Path(entry.stp_filename), export_look_up_table=export_look_up_tables)
```
Shimwell, line 167:
```suggestion
            entry.export_stp(Path(output_folder) / Path(entry.stp_filename))
```

paramak/shape.py
```diff
@@ -310,19 +314,22 @@ def export_stl(self, filename, tolerance=0.001):
 
         return str(Pfilename)
 
-    def export_stp(self, filename=None):
+    def export_stp(self, filename=None, export_look_up_table=False):
```
Shimwell, line 317:
```suggestion
    def export_stp(self, filename=None:
```

paramak/shape.py
```diff
@@ -310,19 +314,22 @@ def export_stl(self, filename, tolerance=0.001):
...
-        filename is not provided and the shapes 
+        filename is not provided and the shapes
         stp_filename property is not None the stp_filename
         will be used as the export filename
 
         :param filename: the filename of the stp
         :type filename: str
+        :param export_look_up_table: if set to True, the look up table
+            will be exported. Defaults to False.
+        :type filename: bool
```
Shimwell, line 329:
```suggestion
```

paramak/shape.py
```diff
@@ -331,15 +338,43 @@ def export_stp(self, filename=None):
...
-        elif self.stp_filename != None:
+        elif self.stp_filename is not None:
             Pfilename = Path(self.stp_filename)
 
         with open(Pfilename, "w") as f:
             exporters.exportShape(self.solid, "STEP", f)
         print("Saved file as ", Pfilename)
 
+        if export_look_up_table:
+            self.export_look_up_table(filename=Pfilename.with_suffix(".json"))
```
Shimwell, line 349:
```suggestion
```

RemiTheWarrior:
@Shimwell maybe it'd be better if we start calling this "physical groups" instead of "look-up tables" as it surely is more explicit and consistent with GMSH for example.

What do you reckon ?

Shimwell:
This all looks good, the name change is nice.

Issue #86, opened 2020-07-17
BUG: BlanketConstantThicknessFP() with full covergae


RemiTheWarrior:
![image](https://user-images.githubusercontent.com/40028739/87774529-eedc1500-c824-11ea-8ff9-37ad45921dd1.png)

How to reproduce:

```python
from paramak import BlanketConstantThicknessFP

my_shape = BlanketConstantThicknessFP(100, start_angle=0, stop_angle=360, rotation_angle=360)
my_shape.solid

```

Shimwell:
fixed by @RemiTheWarrior in the PR #89

RemiTheWarrior:
@Shimwell This issue should be reopened as it hasn't been fixed by PR #89 and won't be in PR #90 ...

PR #87, opened 2020-07-17
Physical groups for BlanketConstantThicknessFP

RemiTheWarrior:
This PR includes a new method for adding physical groups (with names) to `paramak.BlanketConstantThicknessFP()`.
The surfaces are labelled as shown below:
image

## Usage

```python
from paramak import BlanketConstantThicknessFP

my_shape = BlanketConstantThicknessFP(100, start_angle=0, stop_angle=180, rotation_angle=180)

my_shape.export_stp('blanket.stp')
my_shape.export_physical_groups('blanket.json')
```

If `rotation_angle` = 360 then `'left_section'` and `'right_section'` don't exist.

These physical groups can then be read in GMSH  as described in #83.

## Known bugs

-  If `start_angle` = `stop_angle` mod 2*k*pi, `inner_section `and `outer_section ` are not be exported since they don't normally exist. Though `inner_section `and `outer_section ` exist due this bug #86 

## ToDo

- [ ]  Find a way to test this

- [x] Reorder the surfaces names in accordance with internal STEP files surfaces ids in a more generic way (2, 4 or 6 surfaces)

### DIFF
--- a/paramak/parametric_components/blanket_constant_thickness_fp.py
+++ b/paramak/parametric_components/blanket_constant_thickness_fp.py
@@ -246,12 +246,17 @@ class BlanketConstantThicknessFP(RotateMixedShape):
         return points
 
     def create_physical_groups(self):
+        """Creates the physical groups for STP files
+
+        Returns:
+            list: list of dicts containing the physical groups
+        """
+
         def diff_between_angles(a, b):
             c = (b - a) % 360
             if c > 180:
                 c -= 360
             return c
-
         groups = []
         nb_volumes = 1  # only one volume
         nb_surfaces = 2  # inner and outer
--- a/paramak/parametric_components/blanket_constant_thickness_fp.py
+++ b/paramak/parametric_components/blanket_constant_thickness_fp.py
@@ -119,7 +119,8 @@ class BlanketConstantThicknessFP(RotateMixedShape):
 
     @property
     def physical_groups(self):
-        return self.create_physical_groups()
+        self.create_physical_groups()
+        return self._physical_groups
 
     @physical_groups.setter
     def physical_groups(self, physical_groups):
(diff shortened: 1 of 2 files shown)


Shimwell:
ah i understand the other messages now. Perhaps I can add a getter setter to help here

paramak/parametric_components/blanket_constant_thickness_fp.py
```diff
@@ -114,6 +114,7 @@ def __init__(
             self.offset_from_plasma = offset_from_plasma
         self.num_points = num_points
         self.points = points
+        self.physical_groups = self.create_physical_groups()
```
Shimwell, line 117:
```suggestion
        self.physical_groups = None

    @property
    def physical_groups(self):
        self.create_physical_groups()
        return self._physical_groups

    @physical_groups.setter
    def physical_groups(self, physical_groups):
        self._physical_groups = physical_groups
```

Shimwell:
What do you think, this way the creation of physical groups is only called when the property is accessed instead of on the initialization. This also helps with making sure the groups are up to date if the model is changed

Shimwell:
ps the CI tests are failing due to a gitaction we are trying to implement so this can be ignored for now. We can run tests locally

RemiTheWarrior:
@Shimwell sounds super! But the CI don't run the test suite do they ? only the deployment not CircleCI ?

Shimwell:
> @Shimwell sounds super! But the CI don't run the test suite do they ? only the deployment not CircleCI ?

The CI use to run the test suite yes and it will be back up and running again. but it needs a bit of tinkering to get it running tests again like it use to. John has made a good start and is looking into getting it running again next week

Shimwell:
I think this is ready to merge, it doesn't interfere with anything else which is great

RemiTheWarrior:
@Shimwell I think now it's a bit more ready. Still don't know what tests to write....

billingsley-john:
> > @Shimwell sounds super! But the CI don't run the test suite do they ? only the deployment not CircleCI ?
> 
> The CI use to run the test suite yes and it will be back up and running again. but it needs a bit of tinkering to get it running tests again like it use to. John has made a good start and is looking into getting it running again next week

FYI - The automated PYPI build is failing for these pull requests because GitHub secrets are not passed to workflows that are triggered by a pull request from a fork

PR #91, opened 2020-07-17
added expanding blanket to fill divertor gap

Shimwell:
In cases were there is a large distance between the center column and the plasma the divertor was very large. This is not ideal as it removes space from the blanket and results in a bias for parameters studies

So this PR adds radial build options to the divertor with the argument ```divertor_radial_thickness```.

However this introduces a new problem ...

The ```divertor_radial_thickness``` value now controls the dimensions of the divertor and in the case that is smaller than the gap between the blanket arc (which stops at the plasma peak) and the center column there will be a gap where neutrons would leave the geometry.

To solve this new problem the blanket (firstwall and rear wall as well) now expands horizontally to plug the gap

The diagram shows this situation and colours in the extra blanket, fw and rear wall

Screenshot from 2020-07-17 20-28-04

# ToDo

This just leaves one more case to solve which is the case where the ```divertor_radial_thickness```  value is bigger than the gap provided. Another PR will fix this issue.


Shimwell:
This PR appears to also be triggering a PYPI build and failing due to the version number

Shimwell:
tests have been run locally and they pass

Shimwell:
Added some more tests

PR #102, opened 2020-07-21
updated docstrings

billingsley-john:
This PR contains an update to docstrings for some parametric_shapes. The main reason for this PR is to test whether a PR to the develop branch triggers a pip build


billingsley-john:
This PR shows that the PYPI package is still trying to be built

PR #103, opened 2020-07-21
updated docstring

billingsley-john:
PR to test whether the new pip build workflow works. The python-publish.yml file has been updated so that it should only trigger a build on pushes and PRs to the main branch. Before, it was triggering on any PR to any branch


billingsley-john:
This PR shows that the PYPI build workflow has been fixed. PYPI builds are only triggered on PRs and pushes to the main branch. Before, the yml was allowing builds to be triggered on any PR

PR #104, opened 2020-07-21
Add material name setter

billingsley-john:


billingsley-john:
PR which adds a setter to material_tag which raises an error if its length > 27. This ensures the material_tag is suitable for use with DAGMC

PR #107, opened 2020-07-21
New ITER-like divertor

RemiTheWarrior:
This PR includes a new parametric component: ITER-like divertor.

### Usage
```python
from paramak import ITERtypeDivertor

my_shape = ITERtypeDivertor(rotation_angle=180)
my_shape.solid()
```
image

**No-dome version**
```python
from paramak import ITERtypeDivertor

my_shape = ITERtypeDivertor(rotation_angle=180, dome=False)
my_shape.solid()
```
image

This is equivalent to

```python
from paramak import ITERtypeDivertorNoDome

my_shape = ITERtypeDivertorNoDome(rotation_angle=180)
my_shape.solid
```

### Parameters
The parameters are described below
image


Shimwell:
Dome height and dome thickness look quite similar to me, are you sure you want two parameters there instead of one

Shimwell:
Also this is the fanciest parameterized shape so far so congrats

RemiTheWarrior:
> Dome height and dome thickness look quite similar to me, are you sure you want two parameters there instead of one

@Shimwell they are similar, but independent nonetheless. Say you want to vary the dome's height without varying its curvature. Or the opposite, you want to change the curvature but keep the same height --> you need 2 parameters.

See this very precise sketch ! 😃 

image

RemiTheWarrior:
@Shimwell 

Also the functions I wrote (rotate, extend) might be useful for building other shapes in the future. How about we add them in a helpers.py script ?

Shimwell:
Perhaps (rotate, extend) could be in a utils.py in the in the base folder, then we have shape.py, reactor.py and utils.py in that "core" collection. Up to you

Shimwell:
Do you think the domeless and dome divertor should be different parametric shapes?

RemiTheWarrior:
> Do you think the domeless and dome divertor should be different parametric shapes?

@Shimwell I thought it'd be redundant but I can create an additionnal class which calls `ITERtypeDivertor(dome=False)` ?



> Perhaps (rotate, extend) could be in a utils.py in the in the base folder, then we have shape.py, reactor.py and utils.py in that "core" collection. Up to you

Ok I'll do this

RemiTheWarrior:
@Shimwell how about this? this way users have more flexibility : they can either use the `ITERtypeDivertorNoDome `class or `ITERtypeDivertor(dome=False)`

paramak/parametric_components/divertor_ITER.py
```diff
@@ -0,0 +1,294 @@
...
+            the top of vertical targets.
+            Defaults to ((450, -300), (561, -367)).
+        coverages (float, float): coverages (anticlockwise) in degrees of the
+            circular parts of vertical targets. Defaults to (90, 180).
+        radii (float, float): radii (cm) of circular parts of the vertical
+            targets. Defaults to (50, 25).
+        lengths (float, float): leg length (cm) of the vertical targets.
+            Defaults to (78, 87).
+        dome (bool, optional): If set to False, the dome will not be created.
+            Defaults to True.
```
Shimwell, line 20: Do you need these here with the new ITERtypeDivertorNoDome

Shimwell:
Yep that looks great, 2 parametric components is even better than 1

paramak/parametric_components/divertor_ITER.py
```diff
@@ -0,0 +1,294 @@
...
+
+        # casing points
+        casing_points = self.create_casing_points(
+            anchors=(self.IVT_anchor, self.OVT_anchor),
+            C=IVT_points[-1][:2], F=OVT_points[0][:2],
+            targets_lengths=(self.IVT_length, self.OVT_length))
+
+        # append all points
+        points = IVT_points + dome_points + OVT_points + casing_points
+        points.append(points[0])  # don't know why it's needed
```
Shimwell, line 219: this is needed to make a closed face, first and last point are now the same

RemiTheWarrior:
@Shimwell should the docs be added in another PR ?

Shimwell:
> @Shimwell should the docs be added in another PR ?

I don't mind, I do like docs but I don't want that to get in the way of a nice feature

docs/source/paramak.parametric_components.rst
```diff
@@ -100,6 +100,21 @@ DivertorBlock()
...
+    :width: 180px
+    :height: 220px
+    :align: center
+
+
+.. automodule:: paramak.parametric_components.divertor_ITER
+   :members:
+   :show-inheritance:
+
+
```
Shimwell, line 117:
```suggestion
DivertorITERNoDome()
^^^^^^^^^^^^^^^^^^^^^^

.. image:: https://user-images.githubusercontent.com/40028739/88180936-626b9100-cc2e-11ea-92df-1bac68b11e3b.png
    :width: 180px
    :height: 220px
    :align: center
```

RemiTheWarrior: The image isn't right then

RemiTheWarrior: A change was pushed to add the NoDome version

Shimwell:
Unless there are any objections I plan to merge this in tonight

RemiTheWarrior:
> Unless there are any objections I plan to merge this in tonight

@Shimwell ok on my side

Shimwell:
Keen to see if the docs break

RemiTheWarrior:
@Shimwell the images seem a bit squished apparently
image

Shimwell:
the width and height can be changed

    :width: ?px
    :height: ?px

PR #109, opened 2020-07-23
Improved neutronics examples

Shimwell:
Continuing the effort to get materials based on more realistic reactors from references paper.

Added center column shield based on optimal WC / water ratio, assumed same water temp as the divertor for now
Added firstwall material based on HCPB design
Added divertor material based on monoblock demo design

# ToDo
Still missing the inboard tf coil but progressed with center column


Shimwell:
I've found a nice reference for the homogenized materials in all the components now. I've also set plasma parameters to a "typical" ST reactor

PR #110, opened 2020-07-23
New super

Shimwell:
This is just an idea to see what can be done about removing some unused arguments from the user view


billingsley-john:
Looks great @Shimwell. I like the look of this solution.

Shimwell:
What do people think, is this a bet way to do things.

It avoids exposing the users to things they don't want or need.

Any feedback welcome 

@RemiTheWarrior @bielsnohr @DeclanMorbey

RemiTheWarrior:
@Shimwell That's super and it simplifies things for users (as well as the docs!)

I reckon these can still be modified in the super class, right ?

paramak/parametric_components/poloidal_field_coil.py
```diff
@@ -20,31 +20,27 @@ def __init__(
...
-        solid=None,
         stp_filename="PoloidalFieldCoil.stp",
         color=None,
         azimuth_placement_angle=0,
-        points=None,
-        name=None,
+        name='pf_coil',
         material_tag="pf_coil_mat",
         cut=None,
-        hash_value=None,
```
bielsnohr, line 33: The alternative would be to collect these commonly unused arguments into a `**kwargs`, set the defaults in a dictionary below, and loop through the passed argument to modify those in the default dictionary, which is passed to `super.__init__()`

Something like (this could definitely be more pythonic, but you get the idea)
```python
default_dict = {...fill in}

for arg in kwargs:
    if arg in default_dict:
        default_dict[arg] = kwargs[arg]
```

Shimwell: Thanks for the suggestion much appreciated, we can give this a go

Shimwell: We have added kwargs to see how it looks :-)

Shimwell:
This can be closed now as the kwarg was implemented in #115

PR #114, opened 2020-07-25
added draft new plasma

Shimwell:
This PR allows a plasma to be build from the 4 coordinates (top, botton, inner, outer). The 4 plasma shaping parameters are found from these coordinates and then used to build a regular plasma.

This plasma is needed to help make parametric reactors with a vertical build

### DIFF
--- a/paramak/parametric_components/tokamak_plasma_from_points.py
+++ b/paramak/parametric_components/tokamak_plasma_from_points.py
@@ -7,9 +7,9 @@ import scipy
 from paramak import Plasma
 
 
-class Plasma(Plasma):
+class Plasma_from_points(Plasma):
     """Creates a double null tokamak plasma shape that is controlled
-       by 4 shaping parameters.
+       by 3 coordinates.
 
     :param outer_equatorial_x_point: the x value of the outer equatorial of the plasma (cm)
     :type outer_equatorial_x_point: float
@@ -22,7 +22,7 @@ class Plasma(Plasma):
        (outer_equatorial_point, inner_equatorial_point, high_point, low_point)
         as tuples of 2 floats and 4 attributes of the plasma (major_radius, minor_radius,
         elongation, triangularity) as floats
-       
+
     :rtype: paramak shape object
     """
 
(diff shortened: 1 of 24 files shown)


paramak/parametric_components/tokamak_plasma_from_points.py
```diff
@@ -0,0 +1,73 @@
...
+
+import numpy as np
+import scipy
+
+from paramak import Plasma
+
+
+class Plasma(Plasma):
+    """Creates a double null tokamak plasma shape that is controlled
+       by 4 shaping parameters.
```
RemiTheWarrior, line 12: Isn't it 3 parameters here ? Namely inner, outer and high point ? or does the hight point counted as 2 ?

paramak/parametric_components/tokamak_plasma_from_points.py
```diff
@@ -0,0 +1,73 @@
+import math
+from pathlib import Path
+
+import numpy as np
+import scipy
+
+from paramak import Plasma
+
+
+class Plasma(Plasma):
```
RemiTheWarrior, line 10: Can a user choose between this Plasma (from points) and the inherited class Plasma (from parameters) ?

paramak/parametric_components/tokamak_plasma_from_points.py
Shimwell: I was thinking the high and low point are the same, but I guess this is not always true

paramak/parametric_components/tokamak_plasma_from_points.py
Shimwell: I need to finish, this PR but I like your plan of having one Plasma then two others plasma_form_points and plasma_from_parameters. I should rename this one

paramak/parametric_components/tokamak_plasma_from_points.py
RemiTheWarrior: They are symetrical indeed (in the case of an up-down symmetrical plasma). How do you wish to take vertical displacement into account ?

PR #116, opened 2020-07-27
Add hash value

billingsley-john:
PR which adds the hash_value solid construction test to some parametric components, and updates their super().__init__ to use a default_dict


billingsley-john:
wrong branch!

Issue #136, opened 2020-07-30
Improved code coverage


Shimwell:
The code coverage of the test suite has increased 

https://codecov.io/gh/ukaea/paramak/branch/develop

however it is still not at 90%

This is partly because some parts of shape.py are difficult to test in the docker image due to graphics problems.

The other parts are the divertor block that should be either tested or removed

Shimwell:
Code coverage is higher but two tests are still failing so I don't think the higher coverage is being reported https://app.circleci.com/pipelines/github/ukaea/paramak/446/workflows/f8cbe21c-8cc9-410d-9d25-e93c7691d1e1/jobs/459

Shimwell:
Up to 94% now, I think this can be closed for now. Well done everyone

PR #138, opened 2020-07-31
More tests for greener badge

RemiTheWarrior:
This PR is related to #136 and increase the coverage above 90% 🚀 💚 .

Tests have been added to : 
- `Shape`
- `ExtrudeMixedShape`
- `RotateMixedShape`
- `Reactor`
- `BallReactor`

Known errors:
- `export_3d_image` method breaks for all objects (shape or reactor)

ToDo:
- methods like `export_3d_image`, `export_stp`, `export_stl` and all should be tested within the `Shape` test suite and not as currently in the `ExtrudeMixedShape` or `RotateMixedShape` test suites


Shimwell:
Thanks for these, puts us on a nice level.

PR #139, opened 2020-07-31
Fixed the added tests...

RemiTheWarrior:
This PR is the follow-up of #138 which removed some of the tests put at the wrong place


Shimwell:
Thanks for the update

PR #143, opened 2020-08-03
added required scripts for Conda install

Shimwell:
The PR is would help bring a conda install to the project

Currently users have to do conda install for cadquery version 2 and then a pip install for the paramak.

This is not ideal as it can lead to the code bases being installed on different environments

The aim of this PR is to add all the code needed for a Conda install 

This will also help connect to OpenMC which also has a conda install

Currently the task is not ready but this PR shows some of the progress that has been made


Shimwell:
I think this might not be need if we go for conda forge. Will keep looking in to this but might del this PR

Shimwell:
Ah looks like OCP is tricky to add to conda forge so taking another looks at conda cloud

PR #144, opened 2020-08-03
Adding submersion reactor core

Shimwell:
This adds a new parametric reactor (currently just the core) to the code base

This reactor has an inboard blanket unlike the current ball reactor.

It also makes use of the new blanket made by @RemiTheWarrior 

Screenshot from 2020-08-03 08-36-45

# ToDo

Divertors are currently missing from the model
Documentation is also missing
Tests are also missing


Shimwell:
Documentation and tests added :-)

PR #157, opened 2020-08-05
Improved sub reactor

Shimwell:
This PR makes use of the new union and intersect features in the submersion reactor

This also adds TF coils, PF coils, support legs to the model

Also the reactor now raises and error if the plasma high point is outside of the allowable region 

# before

Screenshot from 2020-08-03 16-51-39

# after

Screenshot from 2020-08-04 22-23-51

### DIFF
--- a/docs/source/paramak.parametric_reactors.rst
+++ b/docs/source/paramak.parametric_reactors.rst
@@ -22,11 +22,11 @@ BallReactor()
 SubmersionTokamak()
 ^^^^^^^^^^^^^^^^^^^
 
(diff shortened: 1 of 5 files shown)


Shimwell:
Tests for solids creation, stp file creation and svg file creation have been added for combinations of reactor core with and without the tf and pf coil sets

Shimwell:
Documentation has been updated to include a color diagram, this is the first colour diagram not coloured using the colour scheme we use for the parametric shapes. The reason for this is to allow identification of the different components. If this is approach to colour popular then I can redo the ball reactor as well

Issue #142, opened 2020-08-02
A blanket which has a start and stop thickness


Shimwell:
All of the blankets are Constant Thicknesses which is great but occasionally there is a need for a variable thickness blanket

For example the EU DEMO blanket has a thicker outboard blanket than the inboard blanket https://images.app.goo.gl/5FCPGWVKosSfgJ7Z7

perhaps thickness could allow a list of two thickness and then linear interpolation between the 2 points or a start_thickness and stop_thickness could be used instead

Shimwell:
Perhaps a variable offset from the plasma as well

RemiTheWarrior:
@Shimwell How about a component with thickness distribution along the poloidal section ?

Shimwell:
Yes that sounds even better

RemiTheWarrior:
@Shimwell well in that case this can be achieved by adapting the current blanket and allow the thickness argument to be a callable (a function of poloidal angle).

Shimwell:
this would be super flexible, perhaps a bit harder for the user compared to just two numbers [start_thicknes, stop_thickness]

RemiTheWarrior:
as often: why not do both ?
The thickness parameter can be a float (constant thickness), a tuple of 2 floats (thickness varying linearly between start_thickness and stop_thickness), or a callable (function of angle).

Shimwell:
That would be the ultimate solution.

RemiTheWarrior:
This would be needed if a user would want to make this kind of blanket
![image](https://user-images.githubusercontent.com/40028739/89264953-13612b00-d634-11ea-9709-5a6eafc43956.png)

RemiTheWarrior:
Done in #158 this can be closed

PR #158, opened 2020-08-05
Variable thickness for blanket

RemiTheWarrior:
This PR solves #142. The class `BlanketConstantThicknessFP `has been modified in order to allow blanket thickness to vary along the poloidal direction. The class has therefore been renamed `BlanketFP`.

**Thickness from function of angle**
```python
from paramak import BlanketFP
import numpy as np

def thickness(theta):
    return 100+ 70*np.sin(4*theta)

myShape = BlanketFP(thickness=thickness, start_angle=0, stop_angle=180, rotation_angle=180)
myShape.solid
```
image
```python
from paramak import BlanketFP
import numpy as np

def thickness(theta):
    if theta <= 90:
        return 50
    else:
        return 100

myShape = BlanketFP(thickness=thickness, start_angle=0, stop_angle=180, rotation_angle=180)
myShape.solid
```
image

**Thickness varying linearly**
```python
from paramak import BlanketFP


myShape = BlanketFP(thickness=(1, 100), start_angle=0, stop_angle=180, rotation_angle=180)
myShape.solid
```
image

**Known bug**
- `test_BlanketFP_creation_noplasma` fails due to full coverage + full rotation angle


Shimwell:
This all looks great, happy to merge if you think it is ready @RemiTheWarrior

RemiTheWarrior:
All good to go !

PR #159, opened 2020-08-05
Improved ball reactor

Shimwell:
This updates the ball reactor to make use of recently added intersect and union which allow for some simplification and code reduction.


Shimwell:
Updates to the tests are needed before merging

Issue #154, opened 2020-08-04
New neutronics workflow


Shimwell:
There is an option to have an open source method of making h5m dagmc files.

Here is a demo

https://gist.github.com/pshriwise/52452c37d4b7dd89bdc9374e13c35157

If this were to be implemented we would need to ensure the STL geometry does not overlap on the curves. This could perhaps be done by adding lots of control points to the geometry

RemDelaporteMathurin:
@Shimwell, I think h5m files are now implemented. Can this be closed ?

PR #150, opened 2020-08-04
export_3d_image removed

RemiTheWarrior:
This PR resolves #148 

Coverage : 95%.

PR #162, opened 2020-08-05
Feature reactor export stl

Shimwell:
This PR is aiming to add the export_stl command to reactors so that it can be used in a similar way to the current reactor.export_stp

If this feature is merged before export_3d_image feature is removed in #150 then we can easily render reactor images using the stl files.

This has also become more interesting due to a potential ability to make h5m neutronics geometry directly from stl files as mentioned in #154 . Care must be taken to avoid overlapping geometry but that can be looked at in another PR

ToDo

add tests


Shimwell:
tests have been added

PR #165, opened 2020-08-06
Moving scripts to BlanketFP instead of BlanketConstantThicknessFP

Shimwell:
This changes the use of the old BlanketConstantThicknessFP to the new BlanketFP in a few places

ToDo
Something is wrong with the docs strings in ```paramak/parametric_components/tokamak_plasma.py```


Shimwell:
Fixed the docstring issue in tokamak plasma

Also removed divertor block as it is no longer needed and is not a unique shape

Also added PlasmaFromPoints to documentation

PR #170, opened 2020-08-06
Update shape

billingsley-john:
This PR adds the tet_mesh parameter to the constructor of Shape. This allows tet_mesh to be added to the default_dict of the parametric_shape classes to hide the parameter from the user, but is still accessible.
default_dict has been added to each parametric_shape class to set tet_mesh = None upon construction.
Additional parameters can be added to this default_dict to hide them from users, but keep them accessible.


billingsley-john:
To propagate tet_mesh through to parametric_components, the parameter mus be inherited. Therefore, to remain hidden from the user, it must also be added to the default_dict of each parametric_component class

billingsley-john:
hash_value cannot be hidden from the user in the parametric_shape classes in the same way as in the parametric_component classes

billingsley-john:
What do you think @Shimwell?

PR #176, opened 2020-08-07
Adding body of the Reactor.export_h5m method

pshriwise:
Added in the body of the gist. I took the liberty of moving the attempted PyMOAB import into that method. With the import at the top, the method would return the error "core not found" which is a little esoteric. Now the method will raise an `ImportError` if called when PyMOAB is not installed. 

As I mentioned on slack, PyMOAB isn't in the CI so I haven't been able to verify the test you've added. I'd be happy to add it into the CI image before merging to make sure it works though. Just let me know!

### DIFF
--- a/paramak/reactor.py
+++ b/paramak/reactor.py
@@ -14,10 +14,6 @@ import plotly.graph_objects as go
 import pyrender
 from paramak.shape import Shape
 
-try:
-    from pymoab import core, types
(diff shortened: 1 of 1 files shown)


Shimwell:
Thanks @pshriwise Import error is a nice improvement, I shall merge and get this working on CI

PR #177, opened 2020-08-08
Adding export h5m to reactors

Shimwell:
This adds an export_h5m() method to the reactor class.

The method uses PyMOAB and STL files to create a DAGMC h5m gemoetry file. This does not imprint and merge the surfaces like the Trelis method of creating h5m files demonstrated in the examples. So care must be taken to check for overlaps (perhaps another PR)

The docker image used in CI is being updated to include pymoab

#ToDo
Also currently the export_stl method writes files but perhaps byte io or stream io could be used to avoid the actual file writing


Shimwell:
The docker image is still unable to test as pymoab needs to be installed but tests pass locally

PR #179, opened 2020-08-10
Update physical groups

### LINKED ISSUE #172: Update physical_groups
Like tet_mesh #170 , the attribute physical_groups can be added to the constructor of the shape class and added to the default_dict. This allows physical_groups to be specified upon construction (or default to None), as well as modified after, but 'hides' the parameter from the user.

What do you think @RemiTheWarrior ?

billingsley-john:
PR which allows physical_groups to be specified upon construction, but adds it to the default_dict so that it is hidden from users.
Resolves #172


billingsley-john:
Does this still give expected functionality @RemiTheWarrior ?

billingsley-john:
Implemented in #183

Issue #186, opened 2020-08-11
make_faceted_neutronics_model.py script error


billingsley-john:
find_number_of_volumes_in_each_step_file() function cannot be split over lines, must be one complete line

CURRENT:

geometry_details = find_number_of_volumes_in_each_step_file(
[indent] geometry_details, str(os.path.abspath("."))
)

FIX TODO:

geometry_details = find_number_of_volumes_in_each_step_file(geometry_details, str(os.path.abspath(".")))

Shimwell:
Ah I'm sorry this is probably my fault running the pep8 processor over the whole code base. We can revert these two Trelis scripts

billingsley-john:
Script requires line continuation character ("\") to be explicitly specified for trelis

billingsley-john:
autopep8 removes newline characters, so this is not a viable solution

billingsley-john:
the argument --exclude="make_faceteted_neutronics_model.py" can be used locally to ignore the trelis script, however, this does not seem to work when added to the autopep8.yml file.

billingsley-john:
Resolved across #188, #189, #191, and #193.

yml file required --exclude=make_faceteted_neutronics_model.py (i.e. without " ")

PR #187, opened 2020-08-11
returning to previous version

Shimwell:
@billingsley-john here is one option to fix the trelis python scripts. This PR reverts them to the previous version.

Perhaps one way of helping with #186 but perhaps a bit of a poor solution as it undoes all the pep8 work

For your consideration, feel free to delete, merge or modify this PR


billingsley-john:
Alternative solution provided in #188

billingsley-john:
Alternative solution implemented across #188, #189, #191, #193

PR #190, opened 2020-08-13
added newline character

billingsley-john:
this PR is a test to check whether the file exclusion for autopep8 must be on both branches


billingsley-john:
autopep8 still modified the trelis script. This solution does not work.

Issue #164, opened 2020-08-06
Single null designs


Shimwell:
Would be great to have a parametric reactor that just has a single divertor

This could probably just be a small change to the existing BallReactor and SubmersionTokamak that takes an ['upper'],['lower'] or ['upper', 'lower'] argument and then makes the diverter envelop a little shorter. The divertor envelop is used in conjunction with the intersect feature and th firstwall geometry to make the divertor. The height of the divertor envelope can be reduced so that it only covers the upper or lower part of the firstwall

billingsley-john:
To create a single null design, blanket_fw_rear_wall_envelope must be changed to include only the top or bottom half of the envelope. This cannot be done via BlanketConstantThicknessArcV because the correct spline curvature must be preserved, meaning the 'full' envelope must be constructed. An easy solution would be to simply cut z > 0 or z < 0 with a large box, however, this involves using native CadQuery commands.

A better solution could be to add something like an optional start/stop angle to BlanketConstantThicknessArcV so that segments of this shape can be created. This would allow you to create a shape with the correct spline curvature but extract the upper or lower half to create the envelope for divertor creation. However, I realize that this may take the class further away from its initial purpose as you can lose the vertical 'ends'. (same could be applied to BlanketConstantThicknessArcH)

Shimwell:
Lines 431-439 in https://github.com/ukaea/paramak/blob/develop/paramak/parametric_reactors/ball_reactor.py create the divertor. Perhaps change these so that cylinder made only overlaps lower blanket instead of both upper and lower as it does now

billingsley-john:
Update:
Work has commenced on splitting up the create_components() method in ball_reactor.py so that components are created separately in different methods. This currently requires a lot of returning and passing of objects/attributes to different methods.
At the moment, only models including the optional pf and tf parameters can be built, when they are removed, errors occur as some functions expect the values to be passed.
I think using *args and **kwargs could sort this because any required parameters could be passed, but the function would not error if it didn't get passed the optional pf and tf arguments.

billingsley-john:
Separating create_components() into different methods allows us to rewrite the make_divertor() method for a single_null design, without having to rewrite the whole create_components() method, i.e. we can just inherit the methods for all other component creation.

billingsley-john:
Update:

[...]

PR #196, opened 2020-08-18
Add single null design

billingsley-john:
PR which adds a single null parametric reactor.
The reactor is based on the BallReactor, but takes an additional argument of divertor_position = ["upper", "lower"] which specifies the divertor location.
Adds examples and tests for SingleNullReactor and updates docstrings for BallReactor tests

### DIFF
--- a/examples/example_parametric_reactors/make_all_parametric_reactors.py
+++ b/examples/example_parametric_reactors/make_all_parametric_reactors.py
@@ -49,7 +49,7 @@ def main():
         pf_coil_vertical_thicknesses=[50, 50, 50, 50],
         pf_coil_to_rear_blanket_radial_gap=50,
         pf_coil_to_tf_coil_radial_gap=50,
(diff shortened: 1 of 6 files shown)


billingsley-john:
Resolves #164

RemiTheWarrior:
@billingsley-john does this design have a X point ? it seems you're using the non-null version of the plasma object. If you want to see a sharp x point then you'd have to use `PlasmaBoundaries` object. Depends on what you want :-)

billingsley-john:
You're right, this reactor uses the simple tokamak plasma shape. In the future, we should think about using a more complex plasma, but for now, this is just a simple adaptation of the BallReactor.

billingsley-john:
A different method may be implemented. See #164

billingsley-john:
This particular implementation of a single null reactor using the ball reactor is outdated, a better implementation is in #199

Issue #174, opened 2020-08-06
Adding more plasma plots


Shimwell:
The [make_plasmas.py](https://github.com/ukaea/paramak/blob/develop/examples/example_parametric_components/make_plasmas.py#L134-L141) example script has 4 basic plasma and one more [advanced plasma.](
https://github.com/ukaea/paramak/blob/45443e5696c7f6519ba1972570fb809c33236163/examples/example_parametric_components/make_plasmas.py#L134-L141)

Could this script be upgraded to plot the basic plasma shapes on te left by multiplying their x values by -1 and the advance plasma shapes on the right

3 more advanced plasma also be added to the script using these input [parameters](https://github.com/RemiTheWarrior/plasma-boundaries/blob/master/plasmaboundaries/parameters.py)

@billingsley-john

RemiTheWarrior:
@Shimwell @billingsley-john I can help with that if you want. Care must be taken cause the parameters in plasmaboundaries are rationalised by major_radius.

billingsley-john:
#197 plots basic plasmas on the left and plasmaboundaries plasmas on the right. Major radius and minor radius are required before spheromak equilibrium plasma, NSTX_double_null plasma and NSTX_single_null plasma can be added.

Do you know these values for these plasmas @RemiTheWarrior?

RemiTheWarrior:
@billingsley-john I'm afraid I don't but will dig this up

RemiTheWarrior:
@billingsley-john according to [wikipedia](https://en.wikipedia.org/wiki/National_Spherical_Torus_Experiment), the major radius is 0.85m and minor radius is 0.68m.

billingsley-john:
Resolved in #197

PR #197, opened 2020-08-18
basic plasma shapes now plotted on left

billingsley-john:
PR which plots basic plasmas on the left and plasmaboundaries plasmas on the right

### DIFF
--- a/examples/example_parametric_components/make_plasmas.py
+++ b/examples/example_parametric_components/make_plasmas.py
@@ -213,7 +213,8 @@ def main():
         minor_radius=680,
         triangularity=0.35,
         elongation=2,
-        color="yellow"
+        color="yellow",
+        config="double-null"
     )
 
     NSTX_single_null_plasma_plasmaboundaries = make_plasma_plasmaboundaries(
@@ -223,7 +224,8 @@ def main():
         minor_radius=680,
         triangularity=0.35,
         elongation=2,
-        color="purple"
+        color="purple",
+        config="single-null"
     )
 
     fig = go.Figure()
--- a/examples/example_parametric_components/make_plasmas.py
+++ b/examples/example_parametric_components/make_plasmas.py
@@ -13,7 +13,7 @@ import plotly.graph_objects as go
 def plot_plasma(plasma, name=""):
     """Extract points that make up the plasma and creates a plotly trace"""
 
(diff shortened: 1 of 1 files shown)


billingsley-john:
As described in #174, more plasmas can be added when values for major radius and minor radius are provided

billingsley-john:
Single and double null NSTX plasmas have been added to the make_plasmas script using these input [parameters](https://github.com/RemiTheWarrior/plasma-boundaries/blob/master/plasmaboundaries/parameters.py). However, single null and double null NSTX plasmas look very similar (single null doesn't look like a single null plasma)

RemiTheWarrior:
@billingsley-john could you provide some screenshots ? do you obtain something like this ? 
image

RemiTheWarrior:
> Single and double null NSTX plasmas have been added to the make_plasmas script using these input [parameters](https://github.com/RemiTheWarrior/plasma-boundaries/blob/master/plasmaboundaries/parameters.py). However, single null and double null NSTX plasmas look very similar (single null doesn't look like a single null plasma)

@billingsley-john the default configuration of the `make_plasma_plasmaboundaries` function is `'single-null`. Change this parameter in order to get the correct configuration

examples/example_parametric_components/make_plasmas.py
```diff
@@ -202,12 +202,34 @@ def main():
...
     )
 
+    NSTX_double_null_plasma_plasmaboundaries = make_plasma_plasmaboundaries(
+        name="NSTX_double_null_plasma_plasmaboundaries",
+        A=0,
+        major_radius=850,
+        minor_radius=680,
+        triangularity=0.35,
+        elongation=2,
+        color="yellow"
```
RemiTheWarrior, line 216: in order to obtain a double null configuration, add config='double-null'

billingsley-john:
> > Single and double null NSTX plasmas have been added to the make_plasmas script using these input [parameters](https://github.com/RemiTheWarrior/plasma-boundaries/blob/master/plasmaboundaries/parameters.py). However, single null and double null NSTX plasmas look very similar (single null doesn't look like a single null plasma)
> 
> @billingsley-john the default configuration of the `make_plasma_plasmaboundaries` function is `'single-null`. Change this parameter in order to get the correct configuration

Ah, I see! Thank you very much, making this change seems to work. Thank you very much @RemiTheWarrior

PR #199, opened 2020-08-20
Paramak single null new

billingsley-john:
PR which adds a single null reactor design which inherits from the BallReactor class


billingsley-john:
Some tests are failing sporadically - sometimes the model is not built correctly and returns a Standard_ConstructionError

billingsley-john:
what do you think @Shimwell ?

billingsley-john:
Documentation and updates to repository README are still required

billingsley-john:
There is an error that randomly occurs where the the single null divertor is not created correctly. The unions between the divertor 'cylinder' and the blanket and firstwall sometimes do not work. I am unsure of the origin of this error.

billingsley-john:
BallReactor produces an occasional error in the make_component_cuts() method. The cuts for the firstwall, blanket and blanket_rear_casing are sometimes not performed. This is error is also present in the current develop branch, so it may be better to merge this branch before tackling the problem.

PR #201, opened 2020-08-24
Convert to self

billingsley-john:
This PR changes the ball reactor to use parameters stored as attributes of the instance, i.e. self._attributes, rather than local parameters. This allows all parameters to be passed between the methods via self, rather than in large lists of attributes and return statements


Shimwell:
Nice code reduction

billingsley-john:
Further changes could be made, such as removing create_components() and calling each method individually instead

PR #203, opened 2020-08-24
made example render script

Shimwell:
This moves some code from the reactor.py and shape.py which was incompatible with docker into a stand alone script so that renderings can still be made but the docker based tests can also pass


Shimwell:
I think this can be merged if there are no objections.

a couple of failing tests are not related to this PR

billingsley-john:
Happy for this to be merged

Issue #212, opened 2020-08-27
BUG: blanketFP solid is None if start_angle=stop_angle AND rotation_angle is 360


RemiTheWarrior:
## How to reproduce
```python
# Doesn't work
test_shape = BlanketFP(
    major_radius=300,
    minor_radius=50,
    triangularity=0.5,
    elongation=2,
    thickness=200,
    stop_angle=360,
    start_angle=0,
)
assert test_shape.solid is not None
# Works
test_shape = BlanketFP(
    major_radius=300,
    minor_radius=50,
    triangularity=0.5,
    elongation=2,
    thickness=200,
    stop_angle=340,
    start_angle=0,
)
assert test_shape.solid is not None
test_shape = BlanketFP(
    major_radius=300,
    minor_radius=50,
    triangularity=0.5,
    elongation=2,
    thickness=200,
    stop_angle=360,
    start_angle=0,
    rotation_angle=180
)
assert test_shape.solid is not None

```

RemiTheWarrior:
@Shimwell @billingsley-john This bug doesn't come up if the connections between points isn't spline but "straight". But I haven't been able to fix anything ...

RemiTheWarrior:
Update: It seems to be due to the fact that spline connections doesn't pass through the exact coordinates of the points making some surfaces to overlap somehow.

RemiTheWarrior:
Fixed in #218

PR #218, opened 2020-09-01
Bug fixing and improved coverage

RemiTheWarrior:
This PR fixes all the failing tests in CI and includes:

Major:
- a workaround for #212 by raising an error if users `rotation_angle=360` and full coverage to `BlanketFP`.
- a fix for the test of `Shape()` when start point is end point

Minor: 
- improved code coverage
- minor refactoring

Coverage: 94%


Shimwell:
Nice fix there, one step closer to having another dev to main merge

Issue #216, opened 2020-09-01
BallReactor construction error


billingsley-john:
BallReactor has intermittent construction error, either defined as StandardConstructionError or a 'NoneType' error. Error can be reproduced both locally and within the openmcworkshop/paramak_dependencies docker image

Shimwell:
Thanks for trying the docker environment.
Wondering if this is related to the use of circles that was also causing trouble in the Trelis workflow. We could replace the use of blankets from arcs with the better blanket geometry that uses splines

billingsley-john:
This issue seems to be related to rotation_angle. Tests pass when models are constructed with rotation_angle < 360, but sometimes fail when rotation_angle = 360.
This looks like an internal CadQuery error.
Issue #212 may also be related to this

billingsley-john:
Error is also present for the SubmersionTokamak. This was not picked up earlier as all tests had rotation_angles < 360

Shimwell:
Is it possible to isolate this to a single component.

billingsley-john:
In the case of the BallReactor, it is always when constructing the firstwall, blanket or blanket rear casing. Traceback indicates the error occurs when components are fused or cut

RemiTheWarrior:
@billingsley-john I also found out that if the connections between points (in the case of #212 for the BlanketFP) is `"straight"` everywhere and without `"spline"` connections, then this doesn't appear.

This would indicate it's due to a mismatch caused by the fact that spline connections are not exact.

RemDelaporteMathurin:
is this related to #445 ?

Issue #219, opened 2020-09-01
circle ci builds not triggered from forks


billingsley-john:
Circle Ci builds are not being triggered from forks. This could be because the forks are outside the UKAEA organisation

PR #220, opened 2020-09-01
updated rotation angle

billingsley-john:
updates test_BallReactor.py script to use rotation_angle=359. This prevents the construction error outlined in #216. This is a temporary fix until #216 is resolved


billingsley-john:
@RemiTheWarrior this PR triggered a ci build which failed because it could not pull the image. In reference to #219

PR #221, opened 2020-09-01
added new dockerimage for ci and to docs

Shimwell:
This updates the circle ci to use the latest docker image and adds a few lines to the docs on the use of docker images


Shimwell:
Looks like this gets the tests running again :-)

PR #223, opened 2020-09-01
Add rotation angle warning

billingsley-john:
PR which prints warning message if BallReactor or SubmersionTokamak (or single null versions) rotation_angle = 360. Also updates submersion reactor tests to use rotation_angle=359.


billingsley-john:
@Shimwell what do you think? Would it be better to be a warning or raise an error?

PR #227, opened 2020-09-02
updated warning to use warnings module

billingsley-john:


billingsley-john:
Test for warning needs to be added

PR #236, opened 2020-09-09
Move hash

billingsley-john:
## Proposed changes

PR which moves the hash_value calculation method into shape.py which is inherited by all parametric shapes. This removes a lot of duplicate code from each parametric shape class as the hash_value calculation is the same in each case.

## Types of changes

- [x] New tests
- [x] Refactoring

## Checklist

- [x] Pep8 applied


billingsley-john:
This change also allows tests to be removed from each parametric shape as the method only has to be tested once as part of shape.py

billingsley-john:
CircleCI tests are not working because i think this branch points to the incorrect docker image for running the tests

paramak/parametric_shapes/extruded_circle_shape.py
```diff
@@ -51,6 +47,7 @@ def __init__(
         intersect=None,
         union=None,
         material_tag=None,
+        hash_value=None,
```
Shimwell, line 50: Wondering if it can be removed from here

```suggestion
```

Shimwell:
> This change also allows tests to be removed from each parametric shape as the method only has to be tested once as part of shape.py

But don't we still need to test that is propagates up to the parametric shapes

Shimwell:
> CircleCI tests are not working because i think this branch points to the incorrect docker image for running the tests

ah yes that might be my fault. I tried to change the image from opencmcworkshop/openmc_nndc to opencmcworkshop/openmc 
but I thought I had changed ti back

sorry about that

Shimwell:
Do we have to include the hash_value in the class __init__ as this exposes it to the users. 
Perhaps it could be passed along with other **kwargs when needed and if it is not found in the **kwargs dict it can be set to none

Shimwell:
lots of great code reduction

billingsley-john:
> > This change also allows tests to be removed from each parametric shape as the method only has to be tested once as part of shape.py
> 
> But don't we still need to test that is propagates up to the parametric shapes @billingsley-john

Apologies, I forgot to mention that this has been tested. The parametric shape solid is only reconstructed when changes to its parameters have been made. The hash_value is only updated when .solid is called.

billingsley-john:
> Do we have to include the hash_value in the class **init** as this exposes it to the users.
> Perhaps it could be passed along with other **kwargs when needed and if it is not found in the **kwargs dict it can be set to none

Happy to implement something like this if we think users are likely to want to specify hash_values upon construction. What do you think?

PR #238, opened 2020-09-09
moved hash_value to default_dict

billingsley-john:
PR to move the hash_value parameter to default dict. This removes hash_value from the init constructor, but allows it to be passed as a kwarg.
See #236 for further details.
What do you think @Shimwell ?


Shimwell:
Thanks, this looks good

Shimwell:
feel free to merge

Issue #253, opened 2020-09-16
parametric component points


billingsley-john:
Some parametric components still have last point=first point in their find_points() methods. These can all be removed.

billingsley-john:
This hasn't been picked up because the parametric_component classes have their own points setter, rather than inheriting the points setter from shape.py

billingsley-john:
Some parametric_components still require last point=first point as they overwrite the setter which automatically adds the final point

Shimwell:
Well spotted

billingsley-john:
The points getter and setter should not be overwritten in the parametric component classes. Instead, self.find_points() should be called immediately after the 'parametric parameters' (such as height, inner_radius etc) are set. I.e. self.find_points() does not have to be called as part of the points setter, meaning we can inherit the points getter/setter pair instead and make use of the points checks defined in shape.py

billingsley-john:
Resolved in #254

PR #254, opened 2020-09-16
Parametric component points

billingsley-john:
## Proposed changes

PR which removes the points getter/setter pair specified for each parametric component which overwrites the getter/setter pair in shape.py. As mentioned in #253, find_points() can be called directly outside a getter/setter and the points are checked for errors. 

## Types of changes

- [ ] New tests
- [x] Code refactoring

## Checklist

- [x] Pep8 applied

## Further comments

A similar solution could be implemented for parametric components which overwrite the azimuth_placement_angle getter/setter pair when a method is called to calculate these angles.

### DIFF
--- a/examples/example_parametric_components/make_all_parametric_components.py
+++ b/examples/example_parametric_components/make_all_parametric_components.py
@@ -112,8 +112,6 @@ def main():
     )
     all_components.append(component)
 
-    #
-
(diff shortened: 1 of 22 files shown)


Shimwell:
This is great stuff, code reduction is the super. Thanks for this @billingsley-john

PR #262, opened 2020-09-17
moved cut union intersect into shape

billingsley-john:
## Proposed changes

Code refactoring to move cut, union and intersect getter/setter pairs into shape.py and having each parametric shape inherit. This removes the need for boolean method getter/setter pairs in each parametric shape.
Ideally, the boolean commands for performing the operations should be moved out of create_solid() and into a common method in shape.py which is used in create_solid() by each parametric shape. This also allows the boolean operations to be used by Shape.

## Types of changes

- [ ] New tests
- [x] Code refactoring

## Checklist

- [x] Pep8 applied

## Further comments

No further comments

### DIFF
--- a/paramak/parametric_shapes/extruded_circle_shape.py
+++ b/paramak/parametric_shapes/extruded_circle_shape.py
@@ -68,39 +68,15 @@ class ExtrudeCircleShape(Shape):
             stl_filename=stl_filename,
             azimuth_placement_angle=azimuth_placement_angle,
             workplane=workplane,
+            cut=cut,
+            intersect=intersect,
+            union=union,
             **default_dict
         )
 
-        self.cut = cut
-        self.intersect = intersect
-        self.union = union
         self.radius = radius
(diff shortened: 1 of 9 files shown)


billingsley-john:
Extensive testing must be performed before this branch is merged.
Local tests should be performed and additional tests should be added to the test suite to ensure functionality is as expected.
Boolean operations need to be moved from create_solid() into shape.py

billingsley-john:
Tests need to be re-organised and additional tests need to be added

Shimwell:
Looks good John feel free to merge when you are happy it works.
Submersion reactor might provide a good test as it makes use of boolean opperations

billingsley-john:
The boolean functions work as expected. However, sometimes some components of the single null ball reactor are not created successfully (but an error is not returned). However, I think this is an existing error rather than an error caused by this PR, so will merge and troubleshoot afterwards. Issue will be raised.

PR #289, opened 2020-09-23
adding list to offset and thickness to BlanketFP

Shimwell:
## Proposed changes

Currently BlanketFP accepts a tuple for offset_from_plasma and thickness. This tiny PR adds support for lists (still of length 2). Lists behave the same as tuples in this case and I think lists are generally more used by users.

## Types of changes

- [ ] Code refactoring
- [ ] New tests

## Checklist

- [ x Pep8 applied


Shimwell:
what do you think @RemiTheWarrior  is this ok with you?

RemiTheWarrior:
@Shimwell you are right this is more flexible !

PR #293, opened 2020-09-24
Fix azimuthal placement

billingsley-john:
## Proposed changes

PR which allows the number of inner tf coils to be controlled by the azimuth_placement_angle parameter.
if number_of_coils = value and azimuth_placement_angle = None: value number of tf coils spaced evenly over 360 degrees.
if number_of_coils = value and azimuth_placement_angle = different value: error raised. (both values must be the same).
if number_of_coils = None and azimuth_placement_angle = None: error raised. (at least one must be specified).
if number_of_coils = None and azimuth_placement_angle = value: coils placed at values specified.

## Types of changes

- [x] Code refactoring
- [ ] New tests

## Checklist

- [ ] Pep8 applied

## Further comments

No further comments

### DIFF
--- a/paramak/parametric_components/inner_tf_coils_circular.py
+++ b/paramak/parametric_components/inner_tf_coils_circular.py
@@ -40,13 +40,13 @@ class InnerTfCoilsCircular(ExtrudeMixedShape):
         height,
         inner_radius,
         outer_radius,
-        number_of_coils,
         gap_size,
         distance=None,
         stp_filename="InnerTfCoilsCircular.stp",
         stl_filename="InnerTfCoilsCircular.stl",
         color=(0.5, 0.5, 0.5),
-        azimuth_placement_angle=0,
+        azimuth_placement_angle=None,
+        number_of_coils=None,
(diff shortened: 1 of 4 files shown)


billingsley-john:
what do you think of this solution @Shimwell ?

billingsley-john:
ToroidalFieldCoilPrincetonD also has the same issue

billingsley-john:
At this stage, users can change shape.number_of_coils after construction, but this does not change shape.azimuth_placement_angle because find_azimuth_placement_angle() method is only called upon initial construction. Likewise, changing shape.azimuth_placement_angle after initial construction does not change shape.number_of_coils.
Instead, find_azimuth_placement_angle() should be called upon each solid construction to ensure shape.number_of_coils and self.azimuth_placement_angle are kept up to date with each other. This justifies the use of a method rather than doing checks in the setters of each parameter as performing checks in the setter would prevent the parameters ever being changed as the parameters can only be updated one at a time, meaning an error would always be returned

Shimwell:
Tricky stuff, can a modified hash method help here

PR #298, opened 2020-09-25
fix solid reconstruction after attribute alteration

billingsley-john:
## Proposed changes

In a previous PR, the points setter was removed from all parametric components in an attempt to reduce the total number of lines of code. This did not affect initial construction as find_points() was called explicitly upon construction. However, find_points() was no longer being called upon reconstruction of the solid meaning that attributes could be changed after initial construction, but the points used to construct the shape were not affected, meaning the changes were useless.
This PR adds a conditional call to find_points() to the points setter in Shape which is inherited by all parametric shapes and components. The call is conditional meaning only classes which have the a find_points() method will call this method, allowing the points to be re-calculated correctly after attribute changes. Classes without a find_points() method remain unaffected.

More extensive tests should be added to the test suite to ensure changes to attributes after initial construction are propagated correctly to the solid.

## Types of changes

- [ ] Code refactoring
- [ ] New tests

## Checklist

- [x] Pep8 applied

## Further comments

No further comments

### DIFF
--- a/paramak/parametric_components/blanket_constant_thickness_arc_h.py
+++ b/paramak/parametric_components/blanket_constant_thickness_arc_h.py
@@ -89,7 +89,14 @@ class BlanketConstantThicknessArcH(RotateMixedShape):
         self.inner_mid_point = inner_mid_point
         self.thickness = thickness
 
+    @property
+    def points(self):
         self.find_points()
+        return self._points
+
+    @points.setter
+    def points(self, points):
+        self._points = points
 
     def find_points(self):
 
--- a/paramak/parametric_components/blanket_constant_thickness_arc_h.py
+++ b/paramak/parametric_components/blanket_constant_thickness_arc_h.py
@@ -89,15 +89,6 @@ class BlanketConstantThicknessArcH(RotateMixedShape):
         self.inner_mid_point = inner_mid_point
         self.thickness = thickness
 
-    @property
-    def points(self):
-        self.find_points()
-        return self._points
-
-    @points.setter
-    def points(self, points):
-        self._points = points
-
     def find_points(self):
 
         self.points = [
@@ -118,6 +109,5 @@ class BlanketConstantThicknessArcH(RotateMixedShape):
                 self.inner_upper_point[0] + abs(self.thickness),
                 self.inner_upper_point[1],
                 "straight",
-            ),
-            (self.inner_upper_point[0], self.inner_upper_point[1])
+            )
         ]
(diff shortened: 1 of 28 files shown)


RemiTheWarrior:
@billingsley-john Just a quick thought, if the class inherits the methods of the inherited class, can these setters be placed at the `Shape()` class level to avoid repetition ?

billingsley-john:
After further troubleshooting, changing the attributes of inner_radius, outer_radius

> @billingsley-john Just a quick thought, if the class inherits the methods of the inherited class, can these setters be placed at the `Shape()` class level to avoid repetition ?

I'm not sure that will work as find_points() only exists in the parametric_component classes. Think this would cause an error when using the parametric_shape classes or Shape class directly as the method will not be found

RemiTheWarrior:
> > @billingsley-john Just a quick thought, if the class inherits the methods of the inherited class, can these setters be placed at the `Shape()` class level to avoid repetition ?
> 
> I'm not sure that will work as find_points() only exists in the parametric_component classes. Think this would cause an error when using the parametric_shape classes or Shape class directly as the method will not be found

True. How about putting a dummy method find_point() in the Shape class then ? I just find it weird that one should repeatedly write the setter each time a new class is created since it's generic

billingsley-john:
> > > @billingsley-john Just a quick thought, if the class inherits the methods of the inherited class, can these setters be placed at the `Shape()` class level to avoid repetition ?
> > 
> > 
> > I'm not sure that will work as find_points() only exists in the parametric_component classes. Think this would cause an error when using the parametric_shape classes or Shape class directly as the method will not be found
> 
> True. How about putting a dummy method find_point() in the Shape class then ? I just find it weird that one should repeatedly write the setter each time a new class is created since it's generic

Yes, that is a possible. In fact, this would offer some benefit as it means the final point for each parametric component would be automatically added to self.points as the points setter in Shape would be inherited. At the moment, final point=first point for each parametric component must be specified explicitly as the points setter becomes overwritten.

billingsley-john:
Added conditional call to find_points() in points getter of Shape. Removes need to redefine points getter/setter in parametric components and automatically adds endpoint, and leaves classes with no find_points() method unaffected

billingsley-john:
Currently, the solid is only initially constructed on the first explicit call of .solid. Therefore, any checks in methods such as find_points() or create_solid() are only performed once .solid is called - they are NOT performed upon construction. This is not a major issue, however, it may be good to include self.find_points() in the initial construction of parametric components which would allow these checks to be performed immediately upon construction instead of waiting for the .solid to be called.

PR #333, opened 2020-10-05
added action for dockerhub publish

Shimwell:
## Proposed changes

This adds a github actions that triggers on changes to the main branch.

The action will build and upload a new docker image to ukaea dockerhub with the tag paramak:latest

It makes use of this https://github.com/docker/build-push-action

Assuming this works then there should be some documentation updates to follow. Not sure if the paramak dependency image should also be rebuilt in this way or if that should be kept static?

## Types of changes

- [ ] Code refactoring
- [ ] New tests

## Checklist

- [ ] Pep8 applied


Shimwell:
Just noticed the dockerfile could do with an update so I shall get that done while I'm looking into these things

Issue #364, opened 2020-10-15
Units usage not consistent


Shimwell:
I thought the Paramak was kind of unit less as it just makes geometry with the numbers provided

However when writing to file and using the geometry upsstream in simulations units are required

CADQuery writes STEP files with units of milli meters

Neutronics operates in cm normally

I think the parametric plasma source uses meters

So I'm wondering if we should add units to the docstrings and if those units should be in meters? The actualy numbers might have to be internally scaled to milli meters but any thoughts on this are welcome

Shimwell:
Discussed with Remi

Perhaps addint and attribute for the units eg. 'm', 'cm', 'mm' is the way to go.

It looks like the CAD geometry is always going to be written in mm in the stp file, so I think we should scale the points to be in mm from whatever input the user provides

RemDelaporteMathurin:
@Shimwell when you say "CAD geometry is always going to be written in mm in the stp file" : i thought the stp files were dimensionless ?

As said earlier, internal scaling of the points attribute could do the trick. BUT, if the distance type attribute (eg. radius, height, inner_radius, minor_radius, major_radius, x_point, etc...) aren't scaled accordingly with the scale function, there might be some inconsistencies.

in order not to have this inconsistency between points and these attributes, one would need to add a tailored scaling method to every single parametric_component in the paramak: this is huge.

Moreover, as soon as the attributes of a class are modified, this scaling method should be modified accordingly --> many many tests are required to catch all eventual inconsistencies....

RemDelaporteMathurin:
This shows an example of a scaling method in paramak.Shape() which scales the points attributes coordinates.
https://github.com/RemDelaporteMathurin/paramak/blob/84d7926259ed365e791ea792cd0c6dc807923e34/paramak/shape.py#L400-L411

https://github.com/RemDelaporteMathurin/paramak/blob/84d7926259ed365e791ea792cd0c6dc807923e34/paramak/utils.py#L7-L16

Shimwell:
CADQuery exporter and most exporters write the CAD file in mm and include things like this in the .STP file

```SI_UNIT(.MILLI.,.METRE.)```

I have tried to change these in the STP file (which can be just ASCII text) to ```SI_UNIT(.CENTI.,.METRE.)``` but it appears the either the PPP assumes mm or changing the units of a CAD file is more difficult than simply find and replace

Script for replacing units in a STP file, it didn't appear to work but this is what I tried

```
from tempfile import mkstemp
[...]

PR #365, opened 2020-10-15
Adding units

Shimwell:
## Proposed changes

This PR needs a bit more work but here is a start of a fix for #364 

This adds an arg and attribute of units to the Shape class and an attribute of scaled_points to the shape class

Then there is a scaling_coefficient that finds it's value based on the units

Then scaled_points has a getter that calculates the points based on the current points and scaling_coefficient

I've then made use of these new scaled_points in the 6 basic parametric shape classes and need to do the same in the parametric components. 

Feedback welcome

Todo

Tests
I need to use scaled_points instead of points in the parametric components

Screenshot from 2020-10-15 19-05-44

## Types of changes

- [ ] Code refactoring
- [ ] New tests

## Checklist

- [ ] Pep8 applied

### DIFF
--- a/paramak/shape.py
+++ b/paramak/shape.py
@@ -21,7 +21,7 @@ class Shape:
     classes that inherit from Shape.
 
     Args:
-        points (list of (float, float, float), optional): the x, y, z
+        points (list of (float, float), optional): the x, z
             coordinates of points that make up the shape. Defaults to None.
         name (str, optional): the name of the shape, used in the graph legend
             by export_html. Defaults to None.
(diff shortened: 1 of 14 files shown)


paramak/shape.py
```diff
@@ -96,6 +100,27 @@ def __init__(
...
+        self._scaling_coefficient = value
+
+    @property
+    def scaled_points(self):
+        scaled_points = []
+        for p in self.points:
+            scaled_coordinate = (p[0] * self.scaling_coefficient,
+                                 p[1] * self.scaling_coefficient)
+            scaled_points.append(scaled_coordinate)
+        return scaled_points
```
RemDelaporteMathurin, line 118: I think the potential connection type (ie straight, spline, circle) is being removed in the scaled_points here

Shimwell: ah yes it is not being added to scaled_points. But it is still in points and the Shape class finds the connection type in points instead of scaled points.

Shimwell:
Closing this for now but we might bring it back later

PR #377, opened 2020-10-19
No more error if first and last points are the same

RemDelaporteMathurin:
## Proposed changes

In the following lines of `Shape()` it is clear that having the first and last point identical is required by the code. This PR therefore proposes to get rid of the error raise.

https://github.com/ukaea/paramak/blob/77bdc10a2744fb50062f42342f863bad94a2ba98/paramak/shape.py#L302-L307

## Types of changes

- [x] Code refactoring
- [ ] New tests

## Checklist

- [x] Pep8 applied

## Further comments

### DIFF
--- a/paramak/shape.py
+++ b/paramak/shape.py
@@ -299,11 +299,7 @@ class Shape:
                             3 but not a mixture of 2 and 3")
 
             if len(values) > 1:
-                if values[-1][0] == values[0][0] and values[-1][1] == values[0][1]:
-                    raise ValueError(
-                        "The coordinates of the last and first points are the same."
-                    )
-                else:
+                if values[-1][:2] != values[0][:2]:
                     values.append(values[0])
 
             self._points = values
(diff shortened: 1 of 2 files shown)


billingsley-john:
I like this idea. The only problem I can see is that it would allow users to input things like:

`shape = p.RotateMixedShape(points=[(0, 0, "straight"), (0, 20, "spline"), (20, 20, "spline"), (0, 0, "spline")])`

where the connection on the first and last points are not the same. This does not cause any issues with solid construction, but could be confusing for users.

RemDelaporteMathurin:
> I like this idea. The only problem I can see is that it would allow users to input things like:
> 
> `shape = p.RotateMixedShape(points=[(0, 0, "straight"), (0, 20, "spline"), (20, 20, "spline"), (0, 0, "spline")])`
> 
> where the connection on the first and last points are not the same. This does not cause any issues with solid construction, but could be confusing for users.

You are right. perhaps adding a warning or something ?

Shimwell:
I think that last connection type would be ignored by the code.

Perhaps it is best to just stop the user entering in that last point

RemDelaporteMathurin:
Perhaps just close this PR then :-)

Shimwell:
ok lets close this but keep the branch just in case

PR #382, opened 2020-10-21
Add automatic cutting wedge

billingsley-john:
## Proposed changes

This PR modifies the CuttingWedgeFS shape to use a slightly different method which is more compatible with applying automatic cutting to parametric components.
It also starts adding automatic wedge cutting to the TF coil parametric shapes based on a passed rotation angle.

## Types of changes

What types of changes does your code introduce to the Paramak?
_Put an `x` in the boxes that apply_

- [ ] Bugfix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Code refactoring
- [ ] Documentation Update (if none of the other choices apply)
- [x] New tests

## Checklist

_Put an `x` in the boxes that apply. You can also fill these out after creating the PR. If you're unsure about any of them, don't hesitate to ask. We're here to help! This is simply a reminder of what we are going to look for before merging your code._

- [x] Pep8 applied
- [x] Unit tests pass locally with my changes
- [x] I have added tests that prove my fix is effective or that my feature works
- [ ] I have added necessary documentation (if appropriate)

## Further comments

No further comments

PR #383, opened 2020-10-21
Simplify reactor

Shimwell:
## Proposed changes

This PR makes use of the improved TF coils that are cut to the correct rotation angle.

Due to #382 we are able to remove a few lines from the reactors

I also started implementing the create_solids methods like we have lower down in the code. The geometry is still created on reactor initialization for now. But this has the advantage that the single null reactors can call create_solid and overwrite the make_divertor method

## Types of changes

- [x] Code refactoring
- [ ] New tests

## Checklist

- [ ] Pep8 applied


Shimwell:
Now that #382 has gone in I think this can be merged

Issue #397, opened 2020-10-25
Adding .area, .areas property to shapes


Shimwell:
When making shapes in the paramak we can currently get the volume using the Shape.volume property

This is because we have wrapped the Cadquery operation
https://github.com/ukaea/paramak/blob/87008817a9f89d3bf0308f79f1b14e8fc878502f/paramak/shape.py#L150-L155

We could do with a similar property called .area which returns the total area of the shape. Currently this can be done like this
```python
random_paramak_shape.solid.val().Area()
```
With the new property we would be able to do this
```python
random_paramak_shape.area
```

We could also do with another property called .areas that returns the area of each seperate face. Currently this can be done like this
```python
for face in random_paramak_shape.solid.val().Faces():
    print(face.Area())
```
With the new property we would be able to do this
```python
random_paramak_shape.areas
```

Shimwell:
added this feature in #402

PR #402, opened 2020-10-26
Adding area attributes to shape

Shimwell:
## Proposed changes

Adding attributes area, areas and volumes as described in issues #397

This allows the volumes, total surface area, and surface areas of faces to be found.

Here is an example

Screenshot from 2020-10-26 18-31-56

## Types of changes

- [ ] Code refactoring
- [x] New tests

## Checklist

- [x] Pep8 applied


Shimwell:
This really opens up lots of possibilities regarding the testing of parametric shapes and parametric components.

We can check that the len(Shape.areas) is equal to the number of exspected faces
We can check that compounds (lie the poloidalCoilSets have the expected number of volumes using Shape.Volume
We can check that the area of faces on the parametric component is as expected e.g. there are two equally sized faces

Lots of our tests just check that the object can be created which is a nice start but this would allow a meaningful improvement to be made. I expect that we will find a few cases where the distance needs to be * 0.5 when extrude both = True. We can also get edge length using additional CadQuery methods

PR #436, opened 2020-11-03
30% speed up of the parametric reactors tests

RemDelaporteMathurin:
## Proposed changes

This PR refactores (a bit) the single null submersion reactor and single null ball reactor by removing the initial calls of create_solids() that is already called in super initialisation.

This method is the bottleneck of the computation time and calling it twice only makes it worse.

This tiny change should decrease the total testing time for these tests by 30% ! (from 20min to 14min !!)

## Types of changes

- [x] Code refactoring
- [ ] New tests

## Checklist

- [x] Pep8 applied

## Further comments

### DIFF
--- a/paramak/parametric_reactors/single_null_ball_reactor.py
+++ b/paramak/parametric_reactors/single_null_ball_reactor.py
@@ -71,10 +71,6 @@ class SingleNullBallReactor(paramak.BallReactor):
             number_of_tf_coils=number_of_tf_coils,
             **kwargs)
 
-        self.shapes_and_components = []
-
-        self.create_solids()
-
     @property
     def divertor_position(self):
         return self._divertor_position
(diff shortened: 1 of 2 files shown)


Shimwell:
I think we can't just remove it from the initiation as use cases expect that the Reactor.Solid contains something when requested.

I was half thinking of putting a .solid attribute in the Reactor.py that called self.create_solids() so that the geometry was only constructed when needed. This is similar to the self.create_solids() in Shape.py . Perhaps later hash value could also be used to avoid unnecessary recreation of the same solids in the way we do it in Shape

RemDelaporteMathurin:
@Shimwell the thing is even if it's removed from say SingleNullSubmersionTokamak init, the exact same method is called in super().__init__() (SubmersionTokamak)

In any case it will never be empty

PR #423, opened 2020-10-28
Adding blanket segments to ballreactor

Shimwell:
## Proposed changes

I've have been trying to add segmented blankets to the ball reactor. After some initial progress with the segmentation I've got to the point where filleted edges are needed to make the reactor look better. 

The diagram has a few components hidden to show the reactor blankets and firstwall more clearly.

Screenshot from 2020-11-08 16-37-16

## Types of changes

- [ ] Bugfix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Code refactoring
- [x] Documentation Update (if none of the other choices apply)
- [x] New tests

## Checklist

- [ ] Pep8 applied
- [ ] Unit tests pass locally with my changes
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] I have added necessary documentation (if appropriate)

PR #439, opened 2020-11-04
attempt at hash value for reactors

Shimwell:
## Proposed changes

I can't quite get the hash value for the reactors to work but making a PR just to show progress. The idea is to remove the call to create_solid() from the parametric reactor init section. The create_solids will then be called when attributes change to avoid unnecessarily recalculating the reactor. At the moment the get_hash is reporting a different value each time even when nothing has changed so the PR is not quite working . I've started with the ball reactor but if this can be made to work then it can be rolled out over all the reactors.

## Types of changes

- [ ] Code refactoring
- [ ] New tests

## Checklist

- [ ] Pep8 applied

### DIFF
--- a/paramak/parametric_reactors/ball_reactor.py
+++ b/paramak/parametric_reactors/ball_reactor.py
@@ -122,8 +122,6 @@ class BallReactor(paramak.Reactor):
 
         self.shapes_and_components = []
 
-        self.create_solids()
-
     @property
     def pf_coil_radial_thicknesses(self):
         return self._pf_coil_radial_thicknesses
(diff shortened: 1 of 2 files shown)


paramak/parametric_reactors/ball_reactor.py
```diff
@@ -182,7 +189,7 @@ def _make_plasma(self):
         )
         plasma.create_solid()
```
RemDelaporteMathurin, line 190: Is this useful ?

Shimwell: I guess plasma.create_solid() can be removed

Shimwell: The reason plasma.create_solid is there is because the vertical build need to access the high point of the plasma

RemDelaporteMathurin: I see. Since these points don't need the whole create_solid method to be called (only plasma parameters), how about we create setters/getters for the high points, equatorial points, etc ?

Shimwell: that sounds like the way to go

Shimwell:
This might need rebasing with the develop branch since #423 adds another parametric reactor

Shimwell:
Wondering if the recent merge of #456 means we can we delete this PR ?

What do you think @billingsley-john

billingsley-john:
See #459

PR #449, opened 2020-11-09
updated sweep method to reduce cross-section area variation

billingsley-john:
## Proposed changes

Due to the nature of the sweep method in CadQuery, there was a large variation in the cross-sectional area of swept shapes when sweeping between two end faces along a spline. This PR considerably reduces the variation in cross-section along the spline by placing a surface of the correct dimensions at each point along the spline path, rather than just at the start and end. This is not a perfect solution but is much better than before.

Note, this solution has been implemented in an attempt to create coolant channel pipes with an approximately constant diameter. Forcing areas to be constant at all defined spline points can cause some unexpected results.

## Types of changes

- [x] Code refactoring
- [x] New tests

## Checklist

- [x] Pep8 applied

## Further comments


billingsley-john:
Perhaps this type of sweep should only be performed if specifically required by the user, i.e. the previous sweep without the forcing of a consistent cross-sectional area should be the 'standard' sweep performed.

PR #459, opened 2020-11-11
Add reactor hash value

billingsley-john:
## Proposed changes

PR which implements a reactor_hash_value to allow the conditional reconstruction of all shapes in the reactor object. When .shapes_and_components is called, the current instance of the reactor object is compared with the stored hash value of the currently constructed reactor object. If they are different, the shapes/components are reconstructed and returned. If they are equal, the previously constructed shapes/components are returned.

The hash calculation methods for shapes and reactors have been moved to paramak/utils.py, however, could be moved back into shape.py and reactor.py in the future. Separate hash calculation methods were required for shapes and reactors to prevent looping/recalling the hash check at the reactor level when the shapes/components are constructed (and their own hash values are checked).

This PR also includes some minor changes to other scripts and the test suite which were made made as part of the debugging/testing of this branch.

## Types of changes

- [ ] Code refactoring
- [x] New tests

## Checklist

- [x] Pep8 applied

## Further comments

In the future, get_hash() for shapes could be renamed to get_shape_hash() to be consistent with get_reactor_hash()


billingsley-john:
Tests need to be added

billingsley-john:
Some keys_to_ignore may not be required as we are no longer appending to .shapes_and_components. i.e. may no longer get stuck in an infinite loop of the hash_value constantly changing and re-triggering create_solids()

billingsley-john:
Add a test showing time saving

billingsley-john:
Perhaps the two get hash methods would be more suitable in shape.py and reactor.py to separate them

billingsley-john:
check rotation_angle test fails which were appearing locally

Shimwell (approved):

PR #477, opened 2020-11-12
Allowing floats for casing thicknesses

Shimwell:
## Proposed changes

This small addition allows pf coil case sets to be made with all the same thicknesses. Normally a list was needed for the casing_thickness but now a float is accepted and used as a constant thickness for all cases. 

This was needed so that the reactors can have a default setting of 10 for the case thickness. It  was not possible to use a list for the default setting as the number of coils would not be known.

Hence there will be another PR that adds coil casing to the submersion and the ball reactor

## Types of changes

- [ ] Code refactoring
- [x] New tests

## Checklist

- [ ] Pep8 applied


Shimwell:
the one failed test works locally and looks unrelated to this PR

PR #485, opened 2020-11-13
Add points hash value

RemDelaporteMathurin:
## Proposed changes

For some complex components (with equation solving or optimisers), the `find_points()` method can be very expensive computationnally. This is the case for `ToroidalFieldCoilPrincetonD`, `BlanketFP` and shapes using the points attributes of these shapes.

This same computational issue for `create_solid` has been solved by checking if the instance parameters have changed and running `create_solid` only if they have changed.

This PR proposes to do the same for `find_points`. Therefore another hash value named specific to `find_points `has been added to `Shape()`.

The computational time saved can be estimated with:
```python
import paramak
import time

t = time.time()

magnet = paramak.ToroidalFieldCoilPrincetonD(100, 200, 10, 10, 1)

for i in range(10):
    magnet.inner_points
elapsed = time.time() - t
print(elapsed)
```

**Current behaviour:**
Takes 12.73 s to compute this

**New behaviour:**
Takes 1.26 s to compute this.  --> -90% 

## Types of changes

- [ ] Code refactoring
- [ ] New tests

## Checklist

- [x] Pep8 applied

## Further comments

### DIFF
--- a/paramak/parametric_components/toroidal_field_coil_princeton_d.py
+++ b/paramak/parametric_components/toroidal_field_coil_princeton_d.py
@@ -6,7 +6,7 @@ from scipy import integrate
 from scipy.optimize import minimize
 
 from paramak import ExtrudeMixedShape
-from paramak.utils import calculate_wedge_cut, add_thickness
+from paramak.utils import calculate_wedge_cut, add_thickness, get_hash
 
 
 class ToroidalFieldCoilPrincetonD(ExtrudeMixedShape):
@@ -64,7 +64,7 @@ class ToroidalFieldCoilPrincetonD(ExtrudeMixedShape):
 
     @property
     def inner_points(self):
-        self.find_points()
+        self.points
         return self._inner_points
 
     @inner_points.setter
@@ -73,7 +73,7 @@ class ToroidalFieldCoilPrincetonD(ExtrudeMixedShape):
 
     @property
     def outer_points(self):
-        self.find_points()
+        self.points
         return self._outer_points
 
     @outer_points.setter
--- a/paramak/parametric_components/toroidal_field_coil_princeton_d.py
+++ b/paramak/parametric_components/toroidal_field_coil_princeton_d.py
@@ -6,7 +6,7 @@ from scipy import integrate
 from scipy.optimize import minimize
 
 from paramak import ExtrudeMixedShape
-from paramak.utils import calculate_wedge_cut, add_thickness, get_hash
+from paramak.utils import calculate_wedge_cut, add_thickness
 
 
 class ToroidalFieldCoilPrincetonD(ExtrudeMixedShape):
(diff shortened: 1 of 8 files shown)


billingsley-john:
Unfortunately, this solution breaks the conditional reconstruction of the parametric shapes. Having "_points" in keys_to_ignore means that changing the points of a shape does not cause the solid to be reconstructed. (This is not picked up in the tests because solid reconstruction is not tested explicitly for changing the points of a shape)

RemDelaporteMathurin:
@billingsley-john another get_hash function then ?

billingsley-john:
> @billingsley-john another get_hash function then ?

Yes, that looks like the easiest way

RemDelaporteMathurin:
> > @billingsley-john another get_hash function then ?
> 
> Yes, that looks like the easiest way

What about making the get_hash method generic by adding a ignored_keys argument ?
```python
def get_hash(shape, ignored_keys=[]):

    hash_object = blake2b()
    shape_dict = dict(shape.__dict__)


    for key in ignored_keys:
        if key in shape_dict.keys():
            shape_dict[key] = None

    hash_object.update(str(list(shape_dict.values())).encode("utf-8"))
    value = hash_object.hexdigest()
    return value
```

billingsley-john:
> > > @billingsley-john another get_hash function then ?
> > 
> > 
> > Yes, that looks like the easiest way
> 
> What about making the get_hash method generic by adding a ignored_keys argument ?
> 
> ```python
> def get_hash(shape, ignored_keys=[]):
> 
>     hash_object = blake2b()
>     shape_dict = dict(shape.__dict__)
> 
> 
>     for key in ignored_keys:
>         if key in shape_dict.keys():
>             shape_dict[key] = None
> 
>     hash_object.update(str(list(shape_dict.values())).encode("utf-8"))
>     value = hash_object.hexdigest()
>     return value
> ```

Yes, that could be a good option!

RemDelaporteMathurin:
This could also be used to replace `get_hash_reactor`

Shimwell:
I think this can be merged, those two tests were fixed elsewhere

RemDelaporteMathurin:
> I think this can be merged, those two tests were fixed elsewhere

When running make_parametric_single_null_submersion_reactor.py the following error is raised:
```
Traceback (most recent call last):
  File "examples/example_parametric_reactors/make_parametric_single_null_submersion_reactor.py", line 44, in <module>
    main()
  File "examples/example_parametric_reactors/make_parametric_single_null_submersion_reactor.py", line 38, in main
    my_reactor.export_stp()
  File "/home/paramak/paramak/reactor.py", line 268, in export_stp
    if len(self.stp_filenames) != len(set(self.stp_filenames)):
  File "/home/paramak/paramak/reactor.py", line 47, in stp_filenames
    values.append(shape_or_componet.stp_filename)
AttributeError: 'NoneType' object has no attribute 'stp_filename'
```

Are we sure this is fixed elsewhere ?

PR #505, opened 2020-11-16
Test refactoring

billingsley-john:
## Proposed changes

PR which refactors the parametric shape tests to use a global default shape variable which can be accessed by each test. This prevents significant code repetition.
Also refactors extruded shapes to ensure shapes of the same size (distance) are created when extrude_both = True and False

## Types of changes

- [x] Code refactoring
- [x] New tests

## Checklist

- [x] Pep8 applied

## Further comments

No further comments


tests/test_parametric_shapes/test_ExtrudeMixedShape.py
```diff
@@ -117,108 +85,61 @@ def test_cut_volume(self):
...
+        the correct cut is performed and the volume is correct."""
+
+        self.test_shape.azimuth_placement_angle = [45, 135, 225, 315]
+        test_volume = self.test_shape.volume
+        self.test_shape.rotation_angle = 180
+        assert self.test_shape.volume == pytest.approx(
+            test_volume * 0.5, rel=0.01)
+
+    # def test_extrude_both(self):
+    #     """test that the volume with extrude_both=True is twice as when
```
Shimwell, line 125: perhaps this can be included but it checks that the volumes are similar size

billingsley-john: Yes, this test should be included. At the moment, extrude_both=True generates a solid with double the volume of extrude_both=False meaning somewhere we are still dividing extrusion distance by 2 when we shouldn't

Shimwell: ah thanks for that info. 

If the test is included will it help find the Shapes that are misbehaving. Perhaps this should be included for all extrude shapes

Shimwell:
wondering if the decrease in coverage is due to a smaller selection of shapes being used

tests/test_parametric_shapes/test_ExtrudeMixedShape.py
billingsley-john: This will be fixed on a new branch after this PR has been merged

billingsley-john:
`for vol_90, vol_180 in zip(comps_90_vol, comps_180_vol):
            assert vol_90 == pytest.approx(vol_180 * 0.5, rel=0.1)`

billingsley-john:
Before merging, need to check that no reactors use extrude_both = False which could be causing errors

billingsley-john:
> This assertion was removed from test_SingleNullSubmersionTokamak because it was randomly failing. Therefore, there seems to be a random error occurring with this reactor class.
> `for vol_90, vol_180 in zip(comps_90_vol, comps_180_vol):`
> `assert vol_90 == pytest.approx(vol_180 * 0.5, rel=0.1)`

It was found that an equivalent test on the develop branch also fails randomly, suggesting the problem is already on develop. Therefore, this test refactoring will be implemented and merged, and the problem tackled in a separate PR.

RemDelaporteMathurin:
> This assertion was removed from test_SingleNullSubmersionTokamak because it was randomly failing. Therefore, there seems to be a random error occurring with this reactor class.
> `for vol_90, vol_180 in zip(comps_90_vol, comps_180_vol):`
> `assert vol_90 == pytest.approx(vol_180 * 0.5, rel=0.1)`

I think the failing shape is the plasma (a RotateSplineShape). Any chance #507 solves this ?

billingsley-john:
> > This assertion was removed from test_SingleNullSubmersionTokamak because it was randomly failing. Therefore, there seems to be a random error occurring with this reactor class.
> > `for vol_90, vol_180 in zip(comps_90_vol, comps_180_vol):`
> > `assert vol_90 == pytest.approx(vol_180 * 0.5, rel=0.1)`
> 
> I think the failing shape is the plasma (a RotateSplineShape). Any chance #507 solves this ?

I did some more digging and seems to be one of the BlanketFP shapes which has the incorrect volume - however, this could be due to a problem with a different component. Code coverage has also uncovered a problem. It seems that sometimes def_make_divertor() and def_make_supports() are not called as part of the reactor construction. Issue raised #523

PR #525, opened 2020-11-18
style corrections from code inspector

Shimwell:
## Proposed changes

This is an attempt to improve the code according to code inspector. Mainly snake case implementations to see if the score can be improved.

It looks like the tree tier of code inspector that we have evaluates the default branch. So this set of changes will only impact the score after merging into the main branch.

We had quite a lot of unused import statements in the code. These have been removed and the order has been done according to pep8

## Types of changes

- [x] Code refactoring
- [ ] New tests

## Checklist

- [ ] Pep8 applied


Shimwell:
I think this does not interfere with any open pull requests

PR #544, opened 2020-11-20
added extrusion_start_offset to extruded shapes

billingsley-john:
## Proposed changes

PR which adds an extrusion_start_offset parameter to ExtrudeCircleShape so that the extrusion can be offset in the plane perpendicular to the face. This can be done easily for the ExtrudeCircleShape as the workplane can be offset before the face construction at the parametric_shape level. However, for other extruded shapes, the face is already constructed in super().create_solid() and is not easy to offset afterwards at the parametric_shape level, so the offset will have to be implemented in the base Shape class.

## Types of changes

- [ ] Code refactoring
- [ ] New tests

## Checklist

- [ ] Pep8 applied

## Further comments

### DIFF
--- a/paramak/parametric_shapes/extruded_circle_shape.py
+++ b/paramak/parametric_shapes/extruded_circle_shape.py
@@ -23,6 +23,7 @@ class ExtrudeCircleShape(Shape):
         self,
         distance,
         radius,
+        extrusion_start_offset=0,
         rotation_angle=360,
         extrude_both=True,
         stp_filename="ExtrudeCircleShape.stp",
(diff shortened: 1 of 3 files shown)


Shimwell:
Can we have an SVG image of these shapes with the arguments labeled on mathcha and then into the docs, I noticed that we don't have any SVG diagram for these shapes already so that would be really great

billingsley-john:
> Can we have an SVG image of these shapes with the arguments labeled on mathcha and then into the docs, I noticed that we don't have any SVG diagram for these shapes already so that would be really great

Yes, of course. I will also add more testing.

PR #610, opened 2020-12-01
Fix z_pos and add automatic port cutting to parametric reactors

### LINKED ISSUE #596: Adding fillet radius to rectangular port cutter causes problems with blanket/firstwall construction
When fillet radius of PortCutterRectangular is not none, the outboard blanket/firstwall components in a BallReactor are not constructed correctly, despite the port cutter itself being constructed correctly. Some port cuts are not performed and additional volumes are sometimes created.
Screenshot from 2020-11-30 12-58-21

### LINKED ISSUE #605: z_pos translates in the wrong plane for port cutters
As described

Screenshot from 2020-12-01 14-20-26

billingsley-john:
## Proposed changes

PR which fixes the problem with port positioning. The z_pos parameter has been replaced with a center_point parameter which allows the port cutter faces to be translated in 2 dimensions rather than 1. This ensures placement can be controlled in all workplanes.

PR also attempts to add automatic port cutting to the parametric reactors.
A utility method called perform_port_cutting has been added to the utility.py script to perform this function.
CURRENT STATUS: Works as expected for large ports and small fillet radii.
KNOWN BUGS: For small ports, sometimes not all ports are cut from the outboard blankets and large fillet radii can cause errors.

## Types of changes

- [ ] Code refactoring
- [ ] New tests

## Checklist

- [x] Pep8 applied

## Further comments

No further comments

### DIFF
--- a/paramak/parametric_components/port_cutters_circular.py
+++ b/paramak/parametric_components/port_cutters_circular.py
@@ -51,4 +51,4 @@ class PortCutterCircular(ExtrudeCircleShape):
         self.radius = radius
 
     def find_points(self):
-        self.points = [(0, self.z_pos)]
+        self.points = [(self.z_pos, 0)]
--- a/paramak/parametric_components/port_cutters_circular.py
+++ b/paramak/parametric_components/port_cutters_circular.py
@@ -7,10 +7,9 @@ class PortCutterCircular(ExtrudeCircleShape):
     other components (eg. blanket, vessel,..) in order to create ports.
 
     Args:
-        z_pos (float): Z position (cm) of the port
-        height (float): height (cm) of the port
-        width (float): width (cm) of the port
-        distance (float): extruded distance (cm) of the cutter
+        center_point (tuple of floats): Defaults to (0, 0).
+        radius (float): radius (cm) of port.
+        distance (float): extruded distance (cm) of the cutter.
         stp_filename (str, optional): defaults to "PortCutterCircular.stp".
         stl_filename (str, optional): defaults to "PortCutterCircular.stl".
         name (str, optional): defaults to "circular_port_cutter".
@@ -21,9 +20,9 @@ class PortCutterCircular(ExtrudeCircleShape):
 
     def __init__(
         self,
-        z_pos,
         radius,
         distance,
+        center_point=(0, 0),
         workplane="ZY",
         rotation_axis="Z",
         extrusion_start_offset=1.,
@@ -47,8 +46,8 @@ class PortCutterCircular(ExtrudeCircleShape):
             **kwargs
         )
 
-        self.z_pos = z_pos
+        self.center_point = center_point
         self.radius = radius
 
     def find_points(self):
-        self.points = [(self.z_pos, 0)]
+        self.points = [self.center_point]
--- a/paramak/parametric_components/port_cutters_circular.py
+++ b/paramak/parametric_components/port_cutters_circular.py
@@ -7,7 +7,7 @@ class PortCutterCircular(ExtrudeCircleShape):
     other components (eg. blanket, vessel,..) in order to create ports.
 
     Args:
-        center_point (tuple of floats): Defaults to (0, 0).
+        center_point ((float, float)): Defaults to (0, 0).
         radius (float): radius (cm) of port.
         distance (float): extruded distance (cm) of the cutter.
         stp_filename (str, optional): defaults to "PortCutterCircular.stp".
--- a/paramak/parametric_components/port_cutters_rectangular.py
+++ b/paramak/parametric_components/port_cutters_rectangular.py
@@ -65,7 +65,8 @@ class PortCutterRectangular(ExtrudeStraightShape):
             (self.width / 2, self.height / 2),
             (-self.width / 2, self.height / 2),
         ]
-        points = [(e[0]+self.center_point[0], e[1]+self.center_point[1]) for e in points]
+        points = [(e[0] + self.center_point[0], e[1] +
+                   self.center_point[1]) for e in points]
         self.points = points
 
     def add_fillet(self):
--- a/paramak/parametric_components/port_cutters_rectangular.py
+++ b/paramak/parametric_components/port_cutters_rectangular.py
@@ -56,7 +56,7 @@ class PortCutterRectangular(ExtrudeStraightShape):
         self.height = height
         self.width = width
         self.fillet_radius = fillet_radius
-        self.add_fillet()
+        # self.add_fillet()
 
     def find_points(self):
         points = [
@@ -69,6 +69,13 @@ class PortCutterRectangular(ExtrudeStraightShape):
                    self.center_point[1]) for e in points]
         self.points = points
 
-    def add_fillet(self):
+    def add_fillet(self, solid):
+        if "X" not in self.workplane:
+            filleting_edge = "|X"
+        if "Y" not in self.workplane:
+            filleting_edge = "|Y"
+        if "Z" not in self.workplane:
+            filleting_edge = "|Z"
         if self.fillet_radius is not None and self.fillet_radius != 0:
-            self.solid = self.solid.edges().fillet(self.fillet_radius)
+            solid = solid.edges(filleting_edge).fillet(self.fillet_radius)
+        return solid
--- a/paramak/parametric_components/port_cutters_rectangular.py
+++ b/paramak/parametric_components/port_cutters_rectangular.py
@@ -65,7 +65,7 @@ class PortCutterRectangular(ExtrudeStraightShape):
             (self.width / 2, self.height / 2),
             (-self.width / 2, self.height / 2),
         ]
-        points = [(e[0], e[1] + self.z_pos) for e in points]
+        points = [(e[1] + self.z_pos, e[0]) for e in points]
         self.points = points
 
     def add_fillet(self):
--- a/paramak/parametric_components/port_cutters_rectangular.py
+++ b/paramak/parametric_components/port_cutters_rectangular.py
@@ -7,7 +7,7 @@ class PortCutterRectangular(ExtrudeStraightShape):
     other components (eg. blanket, vessel,..) in order to create ports.
 
     Args:
-        center_point (tuple of floats): Defaults to (0, 0).
+        center_point ((float, float)): Defaults to (0, 0).
         height (float): height (cm) of the port.
         width (float): width (cm) of the port.
         distance (float): extruded distance (cm) of the cutter.
--- a/paramak/parametric_components/port_cutters_rectangular.py
+++ b/paramak/parametric_components/port_cutters_rectangular.py
@@ -71,4 +71,4 @@ class PortCutterRectangular(ExtrudeStraightShape):
 
     def add_fillet(self):
         if self.fillet_radius is not None and self.fillet_radius != 0:
-            self.solid = self.solid.edges('#Z').fillet(self.fillet_radius)
+            self.solid = self.solid.edges().fillet(self.fillet_radius)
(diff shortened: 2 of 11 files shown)


billingsley-john:
UPDATE: With these changes made to the points attributes of these parametric components z_pos translates the port cutters in the expected direction. 
However, perhaps z_pos is a misleading parameter name as direction this parameter translates the solid depends on the workplane specified by the user. I.e. it translates along the axis of the first letter of the workplane. Maybe something like 'vertical_offset' may be more appropriate as a general parameter.

RemDelaporteMathurin:
> UPDATE: With these changes made to the points attributes of these parametric components z_pos translates the port cutters in the expected direction.
> However, perhaps z_pos is a misleading parameter name as direction this parameter translates the solid depends on the workplane specified by the user. I.e. it translates along the axis of the first letter of the workplane. Maybe something like 'vertical_offset' may be more appropriate as a general parameter.

The term "vertical" may also be a bit confusing since some planes are "horizontal". Tricky

billingsley-john:
> > UPDATE: With these changes made to the points attributes of these parametric components z_pos translates the port cutters in the expected direction.
> > However, perhaps z_pos is a misleading parameter name as direction this parameter translates the solid depends on the workplane specified by the user. I.e. it translates along the axis of the first letter of the workplane. Maybe something like 'vertical_offset' may be more appropriate as a general parameter.
> 
> The term "vertical" may also be a bit confusing since some planes are "horizontal". Tricky

Good point. Maybe 'port_offset' could be suitable? Then in the docstrings it can be made clear that this is the offset from the origin in the direction of the first coordinate of the workplane?

RemDelaporteMathurin:
> > > UPDATE: With these changes made to the points attributes of these parametric components z_pos translates the port cutters in the expected direction.
> > > However, perhaps z_pos is a misleading parameter name as direction this parameter translates the solid depends on the workplane specified by the user. I.e. it translates along the axis of the first letter of the workplane. Maybe something like 'vertical_offset' may be more appropriate as a general parameter.
> > 
> > 
> > The term "vertical" may also be a bit confusing since some planes are "horizontal". Tricky
> 
> Good point. Maybe 'port_offset' could be suitable? Then in the docstrings it can be made clear that this is the offset from the origin in the direction of the first coordinate of the workplane?

I wonder if the solution would be to make these ports more generic and ask the user to provide a (float, float) point coordinates in the workplane instead of setting (0, float). This would solve the issue of naming cause we could name this `center_point` for instance.

billingsley-john:
> > > > UPDATE: With these changes made to the points attributes of these parametric components z_pos translates the port cutters in the expected direction.
> > > > However, perhaps z_pos is a misleading parameter name as direction this parameter translates the solid depends on the workplane specified by the user. I.e. it translates along the axis of the first letter of the workplane. Maybe something like 'vertical_offset' may be more appropriate as a general parameter.
> > > 
> > > 
> > > The term "vertical" may also be a bit confusing since some planes are "horizontal". Tricky
> > 
> > 
> > Good point. Maybe 'port_offset' could be suitable? Then in the docstrings it can be made clear that this is the offset from the origin in the direction of the first coordinate of the workplane?
> 
> I wonder if the solution would be to make these ports more generic and ask the user to provide a (float, float) point coordinates in the workplane instead of setting (0, float). This would solve the issue of naming cause we could name this `center_point` for instance.

I like the sound of this solution, it could solve all of these issues. I shall work on implementation.

billingsley-john:
z_pos parameter has been replaced by a center_point attribute to allow center of cutter face to be specified in two dimensions

paramak/parametric_reactors/ball_reactor.py
```diff
@@ -152,6 +165,39 @@ def __init__(
...
+        self.port_height = port_height
+        self.port_width = port_width
+        self.port_distance = port_distance
+        self.port_start_radius = port_start_radius
+        self.port_azimuth_placement_angle = port_azimuth_placement_angle
+        self.port_fillet_radius = port_fillet_radius
+
+        if self.number_of_ports is not None:
+            if self.port_azimuth_placement_angle is None:
+                self.port_azimuth_placement_angle = np.linspace(
```
RemDelaporteMathurin, line 181: numpy isn't imported here, this should fail

billingsley-john:
UPDATE: Architecture of the port cutters is being changed so that the port cutter shape is only filleted on the edges that are strictly necessary. To achieve this, the filleting must be performed before the cutters are placed at each azimuth placement angle otherwise it is impossible to fillet all the required edges without filleting additional edges. This means that the full range of possible fillet radii for the required edge is possible. Before, the additional filleting of incorrect edges limited the fillet radii that were allowed.

paramak/parametric_reactors/ball_reactor.py
billingsley-john: Ah yes, good spot! Will add to branch which will be merged into this branch.

RemDelaporteMathurin: It's weird that this wasn't caught by the tests

paramak/parametric_components/port_cutters_circular.py
```diff
@@ -7,10 +7,9 @@ class PortCutterCircular(ExtrudeCircleShape):
     other components (eg. blanket, vessel,..) in order to create ports.
 
     Args:
-        z_pos (float): Z position (cm) of the port
-        height (float): height (cm) of the port
-        width (float): width (cm) of the port
-        distance (float): extruded distance (cm) of the cutter
+        center_point (tuple of floats): Defaults to (0, 0).
```
RemDelaporteMathurin, line 10:
```suggestion
        center_point ((float, float)): Defaults to (0, 0).
```

I think that for that type of argument this is the correct syntax.

paramak/parametric_components/port_cutters_rectangular.py
```diff
@@ -7,10 +7,10 @@ class PortCutterRectangular(ExtrudeStraightShape):
     other components (eg. blanket, vessel,..) in order to create ports.
 
     Args:
-        z_pos (float): Z position (cm) of the port
-        height (float): height (cm) of the port
-        width (float): width (cm) of the port
-        distance (float): extruded distance (cm) of the cutter
+        center_point (tuple of floats): Defaults to (0, 0).
```
RemDelaporteMathurin, line 10:
```suggestion
        center_point ((float, float)): Defaults to (0, 0).
```

paramak/parametric_reactors/ball_reactor.py
```diff
@@ -152,6 +166,39 @@ def __init__(
...
+                        'number of ports does not equal number of port azimuthal placement angles')
+        else:
+            if self.port_azimuth_placement_angle is not None:
+                self.number_of_ports = len(self.port_azimuth_placement_angle)
+
+        self.port_start_radius = port_start_radius
+
+        if self.port_azimuth_placement_angle is not None:
+            if self.port_start_radius is None:
+                self.port_start_radius = self.major_radius
```
RemDelaporteMathurin, line 200: Could all of that be in setters instead of in the init ?

paramak/utils.py
```diff
@@ -208,6 +208,60 @@ def calculate_wedge_cut(self):
...
+                extrusion_start_offset=self.port_start_radius,
+                fillet_radius=self.port_fillet_radius,
+                azimuth_placement_angle=self.port_azimuth_placement_angle
+            )
+
+        else:
+            raise ValueError('invalid port_type')
+
+        for component in args:
+            component.cut = [port_cutter] + component.cut
```
RemDelaporteMathurin, line 259:
```suggestion
            component.cut.insert(0, port_cutter)
```

RemDelaporteMathurin: Note that in general, using insert is faster than this current method. It's no big deal here though 😄

tests/test_parametric_components/test_PortCutterCircular.py
```diff
@@ -1,19 +1,57 @@
...
+
+        self.test_shape.extrusion_start_offset = 20
+        self.test_shape.azimuth_placement_angle = [0, 90, 180, 270]
+
+        assert self.test_shape.volume == pytest.approx(test_volume * 4)
+
+    def test_absolute_volume(self):
+        """Creates a PortCutterCircular shape and checks that its volume is correct."""
+
+        assert self.test_shape.volume == pytest.approx(math.pi * (20**2) * 300)
```
RemDelaporteMathurin, line 50: Careful here, if another test is added inbetween and changes the instance attributes, this test will fail. To be safe, you can add `self.test_shape.radius= 20`, `self.test_shape.distance = 300` and  `self.test_shape.extrusion_start_offset = 0.0` to ensure that the correct shape is made.

tests/test_parametric_components/test_PortCutterRectangular.py
```diff
@@ -3,19 +3,55 @@
...
+
+        assert self.test_shape.center_point == (0, 0)
+        assert self.test_shape.workplane == "ZY"
+        assert self.test_shape.rotation_axis == "Z"
+        assert self.test_shape.extrusion_start_offset == 1
+        assert self.test_shape.fillet_radius is None
+        assert self.test_shape.stp_filename == "PortCutterRectangular.stp"
+        assert self.test_shape.stl_filename == "PortCutterRectangular.stl"
+        assert self.test_shape.name == "rectangular_port_cutter"
+        assert self.test_shape.material_tag == "rectangular_port_cutter_mat"
```
RemDelaporteMathurin, line 27: As said a bit earlier, these tests may fail if another test which changes the attributes is run before this one. Perhaps instead of using `self.test_shape` a new shape could be created ? or these assertions could be moved to setUp ?

tests/test_parametric_components/test_PortCutterRectangular.py
```diff
@@ -3,19 +3,55 @@
...
+
+        self.test_shape.extrusion_start_offset = 20
+        self.test_shape.azimuth_placement_angle = [0, 90, 180, 270]
+
+        assert self.test_shape.volume == pytest.approx(test_volume * 4)
+
+    def test_absolute_volume(self):
+        """Creates a PortCutterRectangular shape and checks that its volume is correct."""
+
+        assert self.test_shape.volume == pytest.approx(40 * 40 * 300)
```
RemDelaporteMathurin, line 50: same warning here

tests/test_parametric_components/test_PortCutterCircular.py
billingsley-john: In the current setup, any changes to the test_shape made in a test are unique to that particular test - they do not propagate to other tests. I.e. each test imports the shape defined in setUp(), so adding a new test which changes the shape attributes before this test will have no effect.

tests/test_parametric_components/test_PortCutterRectangular.py
billingsley-john: See above

tests/test_parametric_components/test_PortCutterRectangular.py
billingsley-john: See above

tests/test_parametric_components/test_PortCutterCircular.py
RemDelaporteMathurin: Ah ok super then. I thought the issue we had a while ago where stuff was changed but I think that was inside a test wasn't it ? Nevermind 😄

billingsley-john:
number_of_coils is difficult to implement correctly alongside port_azimuth_placement_angle as one impacts the other. Therefore, number_of_coils will be removed as a parameter (it may be added back later)

paramak/parametric_reactors/ball_reactor.py
billingsley-john: The number_of_ports parameter has been removed, removing the need for most of this code. A setter for port_start_radius has been added.

Shimwell:
Quite a lot going on here. Wondering if the ports should be in an inherited version of the base ball and submersion reactors as not everyone will want ports. Would that make it any easier to implement?

billingsley-john:
Features implemented in #714 and #717

PR #629, opened 2020-12-09
added shellFS component and tests

Shimwell:
## Proposed changes

This adds the shell from shape parametric component that we discussed earlier today with @RemDelaporteMathurin and @billingsley-john . 

The final solution is based on a discovery @billingsley-john who found the .shell method. 

The result is a versatile shape with quite a small amount of code.

The shell shape is cut using a WedgeCutterFS() based on the shape passed into the ShellFS()

Unfortunately it fails when trying to offset a certain blanket I'm working on. I believe that is due to the use of splines on the outer part. It works when the blanket is make entirely from straight sections

Also it can fail when the filleted edges make contact each other

TODO
- Docs and images for docs
- add a spline shape to the test
- add a mixed shape to the tests
- add an extrude shape to the tests
- add a hollow shape to the tests

Here are some examples of the shape in use

```python
base_shape = paramak.PoloidalFieldCoil(
    height=50,
    width=50,
    rotation_angle=90,
    center_point=(100,100)
)
casing_int = paramak.ShellFS(shape=base_shape, kind='intersection')
casing_int.export_stp('int.stp')
casing_arc = paramak.ShellFS(shape=base_shape, kind='arc')
casing_arc.export_stp('arc.stp')
```

Screenshot from 2020-12-09 22-43-51

Screenshot from 2020-12-09 22-43-27

## Types of changes

- [ ] Code refactoring
- [x] New tests

## Checklist

- [ ] Pep8 applied

### DIFF
--- a/paramak/parametric_components/shell_fs.py
+++ b/paramak/parametric_components/shell_fs.py
@@ -1,0 +1,80 @@
+
+from paramak import CuttingWedgeFS, Shape
+
+
+class ShellFS(Shape):
+    """Shell From Shape. Creates a shell casing for the provided shape object.
+    Warning some shapes are too complex to shell. The Shell is cut / trimmed
+    with a paramak.CuttingWedgeFS based on the shape passed. This ensures that
+    the shell does does not exceed the rotation angle of the passed shape.
+
+    Args:
+        shape (paramak.Shape): the shape to create a shell / 3D offset around
+        thickness (float): the thickness of the shell casing around the shape
+            (cm). Passed directly to CadQuery.shell(). Defaults to 10.
+        kind (str, optional) : the method used to connect gaps in the resulting
+            shelled shape. Options include 'arc' or 'intersection'. Use 'arc'
+            for rounded edges 'intersection' for sharp edges. Passed directly
+            to CadQuery.shell(). Defaults to intersection.
+        stp_filename (str, optional): defaults to
+            "ShellFS.stp".
+        stl_filename (str, optional): defaults to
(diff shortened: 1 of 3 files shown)


paramak/parametric_components/shell_fs.py
```diff
@@ -0,0 +1,80 @@
...
+        solid = self.shape.solid.shell(
+            thickness=self.thickness,
+            kind=self.kind,
+        )
+
+        if self.shape.rotation_angle < 360:
+            # rotation angle is set by the self.shapes rotation angle
+            cutting_wedge = CuttingWedgeFS(self.shape)
+            # this trims the sold so that it does not exceed the rotation angle
+            solid = solid.cut(cutting_wedge.solid)
```
RemDelaporteMathurin, line 76: As discussed with @billingsley-john yesterday, I think that this additionnal cutting can be avoided by the offset2d method which would give a bit better performances. Do you think it's worth trying ?

RemDelaporteMathurin (commented):
Hi @Shimwell. Nice new feature here. Do you think it's worth trying with the offset2d method ? If i understood well, it's fairly similar but we maybe wouldn't need that additionnal wedge cut at the end.

paramak/parametric_components/shell_fs.py
Shimwell: Yes this might be quicker. The new rotation of the shape to 360 degrees speeds up the shell and makes it more robust as well. I think we might end up with both options one day.

PR #648, opened 2020-12-30
Adding type hinting

Shimwell:
## Proposed changes

Adding a little bit of type hinting to a class and a few functions. Just read this guide https://realpython.com/python-type-checking/ and wanted to give it a go.

I noticed the hover-text in VS code picks up these hints however the more interesting thing for me is the consequences this could have for doc strings.

We could in theory define all the argument and return types using type hinting and have this show up in the docs using a sphinx add on https://github.com/agronholm/sphinx-autodoc-typehints

Just making this tiny PR to start the discussion :-)

When building the docs the type now appears in smaller letters and no longer has the optional tag.

# before
Screenshot from 2020-12-30 17-21-54

# After
Screenshot from 2020-12-30 17-21-42

## Types of changes

- [x] Code refactoring
- [ ] New tests

## Checklist

- [ ] Pep8 applied

### DIFF
--- a/.github/workflows/dockerhub-publish-dependencies.yml
+++ b/.github/workflows/dockerhub-publish-dependencies.yml
@@ -1,0 +1,43 @@
+# This yml file will trigger a Github Action on pushes to the main branch.
+# This Action will build and upload a Docker image to Dockerhub
+# https://github.com/marketplace/actions/publish-docker
+
+name: dockerhub-publish-dependencies
+
+on:
+  push:
+    branches:
+    - develop  
+    - main
+
+jobs:
+  main:
+    runs-on: ubuntu-latest
+    steps:
+      -
+        name: Set up QEMU
+        uses: docker/setup-qemu-action@v1
+      -
+        name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v1
+      -
+        name: Login to DockerHub
+        uses: docker/login-action@v1 
+        with:
+          username: ${{ secrets.DOCKERHUB_USERNAME }}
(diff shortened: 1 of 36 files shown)


RemDelaporteMathurin:
@Shimwell nice stuff overall! Just a shame that the "optionnal" tag disappears it was rather useful.

billingsley-john:
> @Shimwell nice stuff overall! Just a shame that the "optionnal" tag disappears it was rather useful.

The optional tag can be added using this type hinting - trying to figure this out now

billingsley-john:
> > @Shimwell nice stuff overall! Just a shame that the "optionnal" tag disappears it was rather useful.
> 
> The optional tag can be added using this type hinting - trying to figure this out now

`from typing import Optional`

`param: Optional[type]`

can be added. Not optimal formatting but gets the job done.

Screenshot from 2021-01-04 09-47-20

billingsley-john:
Also, unfortunately, it looks like the suggested way to add multiple type options to an argument is to use Union typing.

e.g.
`from typing import Union`
`param: Union[type1, type2]`

But this doesn't look very nice as Union remains as part of the docstring.

Screenshot from 2021-01-04 10-09-16

Shimwell:
> Also, unfortunately, it looks like the suggested way to add multiple type options to an argument is to use Union typing.
> 
> e.g.
> `from typing import Union`
> `param: Union[type1, type2]`
> 
> But this doesn't look very nice as Union remains as part of the docstring.
> 
> For example:
> 
> Screenshot from 2021-01-04 10-09-16

I wonder if readthedocs does this in a nicer way

PR #659, opened 2021-01-06
Develop combine

Shimwell:
## Proposed changes

This brings in recent changes that have been combined

3d mesh tallies
automatic movement of the trelis script for neutronics simulations
type hinting on lots of classes and functions

## Types of changes

- [x] Code refactoring
- [ ] New tests

## Checklist

- [ ] Pep8 applied


Shimwell:
:tada:

paramak/utils.py
```diff
@@ -271,15 +282,44 @@ def get_hash(shape, ignored_keys=[]):
...
+                new_file.write(line.replace(pattern, subst))
+
+    # Copy the file permissions from the old file to the new file
+    copymode(filename, abs_path)
+
+    # Remove original file
+    remove(filename)
+
+    # Move new file
+    move(abs_path, filename)
```
RemDelaporteMathurin, line 320: This utility function isn't tested which makes the diff coverage drop

paramak/shape.py
```diff
@@ -791,7 +795,7 @@ def export_stp(self, filename=None, units='mm', mode='solid'):
                     only accepts 'solid' or 'wire'", self)
 
         if units == 'cm':
-            self._replace(
+            _replace(
```
RemDelaporteMathurin, line 798: this isn't tested

paramak/shape.py
```diff
@@ -1229,6 +1210,6 @@ def export_graveyard(
         """
 
         self.make_graveyard(graveyard_offset=graveyard_offset)
-        self.graveyard.export_stp(Path(filename))
+        new_filename = self.graveyard.export_stp(Path(filename))
 
-        return filename
+        return new_filename
```
RemDelaporteMathurin, line 1215: these added lines aren't tested

paramak/neutronics_utils.py
```diff
@@ -119,3 +124,225 @@ def add_stl_to_moab_core(
...
+            data = data.tolist()
+            error = error.tolist()
+
+            for counter, i in enumerate(data):
+                if math.isnan(i):
+                    data[counter] = 0.
+
+            for counter, i in enumerate(error):
+                if math.isnan(i):
+                    error[counter] = 0.
```
RemDelaporteMathurin, line 278: This line isn't tested

RemDelaporteMathurin, line 278:
```suggestion
           for content in [data, error]:
                for counter, i in enumerate(content):
                    if math.isnan(i):
                        content [counter] = 0.

```

This is a proposition for refactoring. Also this needs testing

paramak/neutronics_utils.py
```diff
@@ -119,3 +124,225 @@ def add_stl_to_moab_core(
...
+        tally_label: str,
+        tally_data,
+        error_data,
+        outfile):
+    try:
+        import vtk
+    except (ImportError, ModuleNotFoundError):
+        msg = "Conversion to VTK requested," \
+            "but the Python VTK module is not installed."
+        raise ImportError(msg)
```
RemDelaporteMathurin, line 306: These lines aren't tested. But then again, I don't know how to test that a module is not found. Would the native ImportError be self-explanatory ? It may be redundant to re-write the error message ? what do you think ?

RemDelaporteMathurin (changes requested):
Very glad that the sweep shapes are fixed! Coverage is a little bit down though. Below are the lines that aren't hit by the tests.

paramak/utils.py
Shimwell: an interesting, I had just moved that from one of the classes to the utils functions. I can added tests

paramak/shape.py
Shimwell: thanks for spotting this, I can add tests for that

paramak/shape.py
Shimwell: thanks for spotting this, I can add tests for that

paramak/neutronics_utils.py
Shimwell: thanks for spotting this, I can add tests for that

paramak/neutronics_utils.py
Shimwell: hummm yes this is trick, I might take this idea to an issue

RemDelaporteMathurin: I think it also applies to other places in the code but can't remember. Wonder what the pullrequest review had to say about this

Shimwell: We can keep track of this additional testing idea over on issue #679

paramak/neutronics_utils.py
Shimwell: Nice refactor there

paramak/neutronics_utils.py
```diff
@@ -119,3 +124,223 @@ def add_stl_to_moab_core(
...
+            )
+            tally = statepoint.get_tally(name=tally.name)
+
+            data = tally.mean[:, 0, 0]
+            error = tally.std_dev[:, 0, 0]
+
+            data = data.tolist()
+            error = error.tolist()
+
+           for content in [data, error]:
```
RemDelaporteMathurin, line 272:
```suggestion
            for content in [data, error]:
```

There's a missing space here causing a bad indentation. Maybe that's my bad...

Shimwell: ah I think i spotted that when fixing the other things and then just noticed this message

Shimwell:
The patch is still low but the other code coverage number is +0.12%

RemDelaporteMathurin:
> The patch is still low but the other code coverage number is +0.12%

Mostly due to neutronics_utils and neutronics model
https://github.com/ukaea/paramak/runs/1704010638

Shimwell:
I think that is about as many tests that can be added without testing the importErrors

paramak/neutronics_utils.py
Shimwell: ah this looks like it is difficult to test, all the statepoint files produced don't have any NAN in them. It is nice to have this protection as just in case a NAN occurs then it is set to 0. But it looks like it doesn't happen very often which is good, but difficult to test

Shimwell:
John has a few more tests in progress #680 which could help raise the coverage a bit more

RemDelaporteMathurin:
> John has a few more tests in progress #680 which could help raise the coverage a bit more

Super! I think the coverage is fine tbh. This can surely be merged in if needed. The question was more on "should we have these try/except imports or not" but this can be answered later :-)

Shimwell:
@billingsley-john that last commit replaced nested tests with discrete unit tests for the test_NeutronicModel

RemDelaporteMathurin (approved):

Shimwell:
Thanks for the tests refactoring John, I think this is ready to merge. I will have a go at the import testing at some point and see if it can be done in a simple manner

Shimwell:
I just added a few more suggestions for the pullrequest.com review. Making steady progress getting through them :+1:

Shimwell:
The pullrequest.com suggestions included changing a few ValueErrors to TypeErrors. I forgot to follow this up in the tests so that is why the tests failed. I think they are all fixed now :-)

RemDelaporteMathurin:
@Shimwell  any idea why the GA CI fails ?

Shimwell:
> @Shimwell any idea why the GA CI fails ?

I think it was due to a small change in the OpenMC DAGMC interface that is fixed now https://github.com/openmc-dev/openmc/pull/1747

I am now rerunning the CI

Shimwell:
Tests are passing on both github actions and circle ci. Happy to merge this one in.

PR #673, opened 2021-01-11
Spectra tally on cells

Shimwell:
## Proposed changes

This adds the ability to tally neutron and photon spectra on the cells.

When defining a model we now have a few cell tallies that are supported (heating, flux, TBR, spectra).

The next tally to add is fast flux which is can be done in a very similar way to spectra but with a two group energy bin and then deletion of the values in the first (thermal) group. Admittedly this is probably not the most efficient way to implement such a tally but good to start somewhere.

Anyway the spectra is by default 709 CCFE energy group so perhaps this should also a user controllable an option in the future.

```
my_model = paramak.NeutronicsModel(
    geometry=self.my_shape,
    source=self.source,
    materials={'center_column_shield_mat': mat},
    cell_tallies=['heating', 'flux', 'TBR', 'spectra'],
    simulation_batches=2,
    simulation_particles_per_batch=2
)
```

Note the new keyword **'spectra'**

## Types of changes

- [ ] Code refactoring
- [x] New tests

## Checklist

- [ ] Pep8 applied

### DIFF
--- a/examples/example_neutronics_simulations/shape_with_spectra_cell_tally.py
+++ b/examples/example_neutronics_simulations/shape_with_spectra_cell_tally.py
@@ -40,9 +40,9 @@ fig = go.Figure()
 
 # this sets the axis titles and range
 fig.update_layout(
-        xaxis={'title': 'Energy (eV)',
-                'range': (0, 14.1e6)},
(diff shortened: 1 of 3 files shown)


Shimwell:
This is just failing on the sweep tasks

PR #704, opened 2021-01-31
Added export html for stp files

Shimwell:
## Proposed changes

This PR adds some helpful functions to the utils

The ```paramak.utils.load_stp_file``` function allows stp files to be loaded and returns the solid and wires.

The ```paramak.utils.export_wire_to_html``` function allows a list of wires to be exported to a html plot.

Reactors can now be viewed from different view_planes

This PR also refactors the various export_html methods in Reactor and Shape classes so that their is less duplication

Screenshot from 2021-01-31 00-12-36

## Types of changes

- [x] Code refactoring
- [ ] New tests

## Checklist

- [ ] Pep8 applied

## Further comments

### DIFF
--- a/paramak/utils.py
+++ b/paramak/utils.py
@@ -1,6 +1,6 @@
 
 import math
-from collections import Iterable
+from collections.abc import Iterable
 from hashlib import blake2b
 from os import fdopen, remove
 from pathlib import Path
@@ -461,7 +461,7 @@ def extract_points_from_edges(
     view_plane: str = 'XZ',
 ):
 
-    if isinstance(edges, list):
+    if isinstance(edges, Iterable):
         list_of_edges = edges
     else:
         list_of_edges = [edges]
--- a/paramak/utils.py
+++ b/paramak/utils.py
@@ -458,7 +458,7 @@ def plotly_trace(
 
 def extract_points_from_edges(
             edges,
-            view_plane='XZ',
+            view_plane: str = 'XZ',
         ):
 
     points = []
@@ -486,7 +486,10 @@ def extract_points_from_edges(
     return points
 
 
-def load_stp_file(filename, scale_factor=1.):
+def load_stp_file(
+    filename: str,
+    scale_factor: float = 1.
+):
     part = importers.importStep(filename).val()
 
     scaled_part = part.scale(scale_factor)
--- a/paramak/utils.py
+++ b/paramak/utils.py
@@ -483,7 +483,7 @@ def extract_points_from_edges(
                         ValueError('view_plane value of ', view_plane,
                                    ' is not supported')
                     vertices.append(correspondance_table[ax])
-                points.append((*vertices))
+                points.append((vertices[0], vertices[1]))
     return points
 
 
--- a/paramak/utils.py
+++ b/paramak/utils.py
@@ -470,20 +470,24 @@ def extract_points_from_edges(
 
     for edge in list_of_edges:
         for vertex in edge.Vertices():
-            correspondance_table = {
-                "X": vertex.X,
-                "Y": vertex.Y,
-                "Z": vertex.Z,
-                "R": math.sqrt(math.pow(vertex.X, 2) + math.pow(vertex.Y, 2))
-            }
-            vertices = []
-            for edge in list_of_edges:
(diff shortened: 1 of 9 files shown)


tests/test_example_shapes.py
```diff
@@ -85,6 +88,50 @@ def test_make_can_reactor_from_points(self):
...
+            "example_shape_from_solid.html",
+            "example_shape_from_stp.html",
+        ]
+        for output_filename in output_filenames:
+            os.system("rm " + output_filename)
+        make_html_diagram_from_stp_file.main()
+        for output_filename in output_filenames:
+            assert Path(output_filename).exists() is True
+            os.system("rm " + output_filename)
+
```
Shimwell, line 104:
```suggestion

    def test_list_of_wires_can_be_exported(self):
        """Checks than a list of wires is an acceptable input
        for export_wire_to_html wires argument.
        """
        
        example_shape = paramak.ExtrudeMixedShape(
            distance=1,
            points=[
                (150, 100, "spline"),
                (140, 75, "spline"),
                (110, 45, "spline"),
            ]
        )

        fig = paramak.utils.export_wire_to_html(
            wires=[example_shape.wire],
            tolerance=0.1,
            view_plane="XY",
            facet_splines=True,
            facet_circles=True,
            filename="example_shape_from_stp.html",
        )
        
        assert fig is not None

```

Shimwell:
I think this is now ready and if there are no objections I can merge this in tomorrow

paramak/utils.py
```diff
@@ -456,34 +456,168 @@ def plotly_trace(
...
+            elif view_plane == 'ZY':
+                points.append((vertex.Z, vertex.Y))
+            elif view_plane == 'ZX':
+                points.append((vertex.Z, vertex.X))
+            elif view_plane == 'RZ':
+                xy = math.pow(vertex.X, 2) + math.pow(vertex.Y, 2)
+                points.append((math.sqrt(xy), vertex.Z))
+            else:
+                raise ValueError('view_plane value of ', view_plane,
+                                 ' is not supported')
```
RemDelaporteMathurin, line 490: This can be refactored so that it's more generic. What do you think about something like that ?

```suggestion
    for edge in list_of_edges:
        for vertex in edge.Vertices():
            correspondance_table = {
                "X": vertex.X,
                "Y": vertex.Y,
                "Z": vertex.Z,
                "R": math.sqrt(math.pow(vertex.X, 2) + math.pow(vertex.Y, 2))
            }
            vertices = []
            for edge in list_of_edges:
                for ax in view_plane:
                    if ax not in correspondance_table.keys():
                        ValueError('view_plane value of ', view_plane,
                                   ' is not supported')
                    vertices.append(correspondance_table[ax])
                points.append((*vertices))
```

RemDelaporteMathurin (changes requested):
A tiny comment for refactoring and future-proof when/if additionnal view_planes are added

paramak/utils.py
Shimwell: that is much more elegant

paramak/utils.py
```diff
@@ -456,34 +456,168 @@ def plotly_trace(
...
-def export_html_stp_file():
+def extract_points_from_edges(
+    edges,
+    view_plane: str = 'XZ',
+):
+
+    if isinstance(edges, list):
+        list_of_edges = edges
+    else:
+        list_of_edges = [edges]
```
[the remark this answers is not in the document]
Shimwell: ah yes, perhaps I should make it accept instances of iterables

paramak/utils.py
Shimwell: This is nice in terms of the eligance but unfortunately it has a few issues

it is not raising a value error when the view_plane is incorrect.
It is also doing a math.sqrt and math.pow for every vertex even if it is not needed. 
The points.append() is not add a tuple but actually trying to append a couple of items (which fails)

So I think I will switch back to the old method, which is perhaps a little easier to understand. Perhaps this is an upgrade for another time

paramak/utils.py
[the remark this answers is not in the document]
Shimwell: Now accepting iterates

PR #710, opened 2021-02-03
added type hints to args and removed from docstring

Shimwell:
## Proposed changes

I am keen to add type hinting to all 12 of the parametric shapes but wanted to check with others if this is a good idea.

So I made a example PR that implements type hinting on one parametric shape and removes the types from the docstring.

If this is acceptable then I can roll this out for all the other 11 shapes

Let me know what you think @RemDelaporteMathurin @billingsley-john 

## Types of changes

- [x] Code refactoring
- [ ] New tests

## Checklist

- [ ] Pep8 applied

### DIFF
--- a/paramak/parametric_shapes/extruded_circle_shape.py
+++ b/paramak/parametric_shapes/extruded_circle_shape.py
@@ -1,4 +1,6 @@
 
+from typing import Optional
+
 import cadquery as cq
 from paramak import Shape
 from paramak.utils import calculate_wedge_cut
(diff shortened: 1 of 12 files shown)


RemDelaporteMathurin:
Hi @Shimwell, could you maybe provide an example of testing these type hints ?
Like is there a way to avoid "human" errors in the arguments types ? Could we add a test that tests that when the indicated type is passed as an argument the class has the correct behaviour ?

Shimwell:
> Hi @Shimwell, could you maybe provide an example of testing these type hints ?
> Like is there a way to avoid "human" errors in the arguments types ? Could we add a test that tests that when the indicated type is passed as an argument the class has the correct behaviour ?

I am hoping that Mypy can be used test the example scripts use the classes with the correct type errors

Not quite got it working yet but I think that would be able to scan our usage examples and check they conform

RemDelaporteMathurin:
> The actual checking of user inputs and returning of errors is probably best done in the @Property setters

I don't quite get what's the added value of this type hinting then. If I understood correctly, it doesn't automate the docstrings (we still have to manually write the types of parameters right ?) and it's not built for testing.

Sure I'm missing something here

Shimwell:
> > The actual checking of user inputs and returning of errors is probably best done in the @Property setters
> 
> I don't quite get what's the added value of this type hinting then. If I understood correctly, it doesn't automate the docstrings (we still have to manually write the types of parameters right ?) and it's not built for testing.
> 
> Sure I'm missing something here



> > The actual checking of user inputs and returning of errors is probably best done in the @Property setters
> 
> I don't quite get what's the added value of this type hinting then. If I understood correctly, it doesn't automate the docstrings (we still have to manually write the types of parameters right ?) and it's not built for testing.
> 
> Sure I'm missing something here

Ah yes sorry I haven't really explained the motivation for this.

In addition to providing some practice pull requests for Katie the type hints have a key advantage over types in the doc strings.

type hints are understood by mypy, this means that we can check that our usage of each class and function is using the correct arg types.

This can be run and it should check our examples folder and our internal usage within the paramak.

For example if we specify that a function accepts strings and we are internally passing a pathlib.Path type then mypy would detect this and then we could either change the function to accept Paths or change the usage to a string.

Further convincing available here https://www.youtube.com/watch?v=S75VVAZmqZ8

Shimwell:
first results from mypy

this was output by running the following commands
```bash
pip install mypy
cd ~/paramak
mypy paramak
```

Screenshot from 2021-02-03 19-33-57
Screenshot from 2021-02-03 19-34-08

Shimwell:
I think this is now ready to merge if everyone is happy.

billingsley-john (approved):
These are some great additions for documenting the code better

Shimwell (commented):
thanks everyone for pitching in on this one

PR #722, opened 2021-02-09
added openmp threads option

billingsley-john:
## Proposed changes

PR which adds an optional openmp_threads argument to .simulate() to control the number of threads OpenMC uses during a simulation.

## Types of changes

- [ ] Code refactoring
- [ ] New tests

## Checklist

- [ ] Pep8 applied

## Further comments

### DIFF
--- a/paramak/parametric_neutronics/neutronics_model.py
+++ b/paramak/parametric_neutronics/neutronics_model.py
@@ -592,7 +592,7 @@ class NeutronicsModel():
 
     def simulate(self, verbose: bool = True, method: str = None,
                  cell_tally_results_filename: str = 'results.json',
-                 openmp_threads: int = None):
+                 threads: int = None):
         """Run the OpenMC simulation. Deletes exisiting simulation output
         (summary.h5) if files exists.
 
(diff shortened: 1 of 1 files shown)


billingsley-john:
What do you think of something like this? @Shimwell @RemDelaporteMathurin

Shimwell:
Looks fine to me, None takes all the threads which is the default

Might need to add some clear doc strings as the default of None which results in all the threads could confuse

Shimwell:
thanks for that, docstring

any thoughts on the testing

could make use of time and check that a 2 core is faster than the 1 core. But I'm not sure if the CI has many cores available.

https://www.programiz.com/python-programming/time

perhaps just local testing is fine for this

billingsley-john:
I shall explore some testing options to see how easy it would be. I agree, timings would be the way to about it

Shimwell (approved):
Looks good, tests for this are optional as there is testing in openmc for threads scaling

Shimwell (approved):
thanks for that change to make it the same as the openmc argument

PR #724, opened 2021-02-09
Pull request review fixes 0.23

Shimwell:
## Proposed changes

Implemented a few more suggestions from the pullrequest.com review
https://github.com/ukaea/paramak/pull/639

We are down to just 40 remaining suggestions :tada: 

## Types of changes

- [x] Code refactoring
- [x] New tests

## Checklist

- [ ] Pep8 applied

### DIFF
--- a/run_tests.sh
+++ b/run_tests.sh
@@ -1,6 +1,6 @@
 #!/bin/bash
 
-pytest tests/test_neutronics_utils.py -v --cov=paramak --cov-append --cov-report term --cov-report xml
+pytest tests/test_neutronics_utils.py -v --cov=paramak --cov-report term --cov-report xml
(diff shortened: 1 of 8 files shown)


run_tests.sh
```diff
@@ -1,3 +1,5 @@
+#!/bin/bash
+
 pytest tests/test_neutronics_utils.py -v --cov=paramak --cov-append --cov-report term --cov-report xml
```
RemDelaporteMathurin, line 3: Just saying, you don't want the --cov-append flag at the very first line. It won't append to anything

Shimwell: Ah nice spot, I will fix that

Shimwell: fixed in 912d0b7

Issue #445, opened 2020-11-06
RotateSplineShape.volume = 0 if rotation_angle=360


billingsley-john:
As described

RemDelaporteMathurin:
If that can help, the following MWE reproduces the issue by creating an empty STL file:

```python
import paramak
points = [
    (1, 0),
    (2, 0),
    (2, 1),
    (1, 1)
]
shape = paramak.RotateSplineShape(points=points)
shape.export_stl("out.stl")
```

Shimwell:
Perhaps we should update the cadquery version we are running

billingsley-john:
RotateCircleShape is also having the same issue

billingsley-john:
Also the same issue for .area and .areas

billingsley-john:
These issues are not necessarily due to a failed solid construction. In some cases, a solid is correctly produce but the area and volume attributes do not return the expected values.

RemDelaporteMathurin:
With the MWE I put up above, it creates a solid but the STP file is empty (0 Ko)

RemDelaporteMathurin:
@billingsley-john @Shimwell here's a MWE for testing/debugging: 
```python
import paramak
import numpy as np

# create a circle distribution
thetas = np.linspace(0, 2*np.pi, 100, endpoint=False)
offset = 1
radius = 1
x, y = offset + radius*(np.cos(thetas) + 1), radius*np.sin(thetas)

points = [[X, Y] for X, Y in zip(x, y)]

# create 3 nearly identical shapes
shape_spline = paramak.RotateSplineShape(
    points=points[:])


shape_circle = paramak.RotateCircleShape(
    radius=radius, points=[[2, 0]])

shape_straight = paramak.RotateStraightShape(
    points=points[:])

# print volumes
print(shape_spline.volume)
print(shape_circle.volume)
print(shape_straight.volume)

# export to STL
shape_circle.export_stl('circle.stl')
shape_spline.export_stl('spline.stl')
shape_straight.export_stl('straight.stl')


```

RemDelaporteMathurin:
I've narrowed it down to 2 problems:

1.  revolve has to be used with clean=False argument (see [this cadquery issue](https://github.com/CadQuery/cadquery/issues/322))
2. the rotate_solid method breaks stuff ... see the MWE below

```python
import cadquery as cq

## works
solid = cq.Workplane("XY").moveTo(30, 30).circle(10).revolve(360, clean=False)
[...]

PR #733, opened 2021-02-17
DEMO reactor

RemDelaporteMathurin:
## Proposed changes

A tiny PR adding an example script. What do you think @Shimwell ?
## Types of changes

- [ ] Code refactoring
- [ ] New tests

## Checklist

- [x] Pep8 applied

## Further comments


shimwell:
Thanks for making this Remi. I think it is just what we needed. Lets move it to a class, add some tests and some docs in a similar way to PR #734

shimwell:
Current model looks like this 

Screenshot from 2021-02-17 19-30-58

shimwell:
I've moved the example in to a class object, added tests and added some documentation.

I think the general shapes need a bit more work (especially that overlap) and the documentation picture needs updating

RemDelaporteMathurin:
Perhaps the TF coils need an additionnal bit of work cause only the cases are described in this paper. Plus there's an overlap with the divertor

shimwell:
> Perhaps the TF coils need an additionnal bit of work cause only the cases are described in this paper. Plus there's an overlap with the divertor

I guess we can even not have coils in the model to begin with

shimwell:
Would you like me to split this reactor up into sections like I've done for the Sparc one?

RemDelaporteMathurin:
> Would you like me to split this reactor up into sections like I've done for the Sparc one?

I can try and have a go at it today. What about the coils though, should we remove them for now ?

shimwell:
> > Would you like me to split this reactor up into sections like I've done for the Sparc one?
> 
> I can try and have a go at it today. What about the coils though, should we remove them for now ?

Yep lets remove the TF coils and just leave the TF cases. Thanks for this Remi, much appreachiated

RemDelaporteMathurin:
> > > Would you like me to split this reactor up into sections like I've done for the Sparc one?
> > 
> > 
> > I can try and have a go at it today. What about the coils though, should we remove them for now ?
> 
> Yep lets remove the TF coils and just leave the TF cases. Thanks for this Remi, much appreachiated

But even the casing is overlapping I'm afraid. How about removing both the coils and casing until a better diagram is found ?

shimwell:
Perhaps we can make the casing using an extrudeMixedShape and some coordinates

shimwell:
@Derek-yfqiu this reactor model is based on a drawing in a published paper and has the curved magnet cases that you were looking for.

RemDelaporteMathurin:
@Shimwell I've pushed some aditionnal shapes, but something doesn't work with the vacuum vessel... can't figure out what.

shimwell:
I think the points might be messed up on the outer section halfway up 
Screenshot from 2021-02-27 13-45-52

RemDelaporteMathurin:
> I think the points might be messed up on the outer section halfway up

Thanks for spotting this, it now works properly: 
image

shimwell:
Looks like the demo reactor is taking a while to run on the CI and timing out. Perhaps we can reduce the number of points to speed up the model

RemDelaporteMathurin:
> Looks like the demo reactor is taking a while to run on the CI and timing out. Perhaps we can reduce the number of points to speed up the model

hm.. that's odd, it takes a few seconds on my computer...

RemDelaporteMathurin:
Reducing the number of tf coils seems to greatly reduce the generation time. Must be due to the increased number of cuts/union operations. Being able to order the boolean operations could help (see #747 )

RemDelaporteMathurin:
The issue doesn't seem to be fixed though. Don't know what to do about it... can someone try and run it locally ?

shimwell:
> The issue doesn't seem to be fixed though. Don't know what to do about it... can someone try and run it locally ?

Yep I will try that next week, we can always split it up into separate methods and test them separately

shimwell:
I can confirm that it runs quickly over here as well

shimwell:
I shall start chopping up the create_solids into separate methods to see if I can isolate the slow section.

shimwell:
The slow parts appear to be the graveyard construction and the tf casing construction.

RemDelaporteMathurin:
> The slow parts appear to be the graveyard construction and the tf casing construction.

Graveyard: I reckon that's because there are really small gaps between some shapes ? perhaps between tf casing and some pf coils ?
Tf casing: I have no idea... this is from points so it cannot be du to the Princeton-D shape computation.

shimwell:
I've speed the make_graveyard by a few seconds but it takes time to find the largest_dimension attribute of each volume. Perhaps we can add an argument that allows the size of the graveyard to be user specified and skip the automatic calculation

shimwell:
When I run the pytests can see that it often uses more than one core and the system time ends up being 44 mins while real time is just 1 min. So I guess it runs fast on our multicore desktops and slowly on the CI

RemDelaporteMathurin:
@Shimwell perhaps removing the cuts in the vacuum vessel (blanket + divertor) would speed up the computation time ? but there will be some small overlaps

shimwell:
> @Shimwell perhaps removing the cuts in the vacuum vessel (blanket + divertor) would speed up the computation time ? but there will be some small overlaps

lets avoid the overlaps for now.

I have a half implemented method of speeding up the graveyard construction. So I hope that will help. Happy to leave this PR for a few days while I tidy things up elsewhere and fix that graveyard problem. Then I shall be back.

RemDelaporteMathurin:
@Shimwell I'm afraid there's something else going on here. When I run the tests locally, test_vessel_construction takes forever. Removing the line `cut=vac_vessel_inner,  # hollow shape` makes it run instantaneously.

This is rather weird...

RemDelaporteMathurin:
I've noticed that some points in the vacuum vessel weren't ordered correctly. I've fixed it.

However, it seems that when trying to do a 360 rotation angle, the following error occurs: 
```
paramak/reactor.py:305: in export_stp
    entry.export_stp(
paramak/shape.py:811: in export_stp
    exporters.export(self.solid, str(path_filename), exportType='STEP')
paramak/shape.py:138: in solid
    self.create_solid()
paramak/parametric_shapes/extruded_mixed_shape.py:99: in create_solid
    solid = self.perform_boolean_operations(solid, wedge_cut=cutting_wedge)
paramak/shape.py:1137: in perform_boolean_operations
    solid = cut_solid(solid, self.cut)
paramak/utils.py:128: in cut_solid
    solid = solid.cut(cutting_solid.solid)
/opt/conda/lib/python3.8/site-packages/cadquery/cq.py:3069: in cut
    newS = newS.clean()
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = <cadquery.occ_impl.shapes.Compound object at 0x7f3f5d14aee0>

    def clean(self: T) -> T:
        """Experimental clean using ShapeUpgrade"""

        upgrader = ShapeUpgrade_UnifySameDomain(self.wrapped, True, True, True)
        upgrader.AllowInternalEdges(False)
>       upgrader.Build()
E       OCP.Standard.Standard_NullObject: BRep_Tool:: TopoDS_Vertex hasn't gp_Pnt

/opt/conda/lib/python3.8/site-packages/cadquery/occ_impl/shapes.py:360: Standard_NullObject
```

I fear this is somehow related to issue #445 because replacing the RotateSplineShape by RotateStraightShape solves the issue.

shimwell:
Nice detective work there Remi. I am not sure what to do about the 360 rotation error, perhaps this is a question for CQ

RemDelaporteMathurin:
@Shimwell I've opened #757 to try and solve this issue. A workaround for now would be to change the TF coils azimuth_placement_angles to avoid that they are placed at zero degree.

RemDelaporteMathurin:
@Shimwell this is ready to be merged I think. However, the diff coverage is low. This is due to neutronics_utils.py, which is weird cause it's unchanged in this PR. Don't know what to do about it

RemDelaporteMathurin:
@Shimwell @billingsley-john  this is now ready for review! :-D

paramak/parametric_reactors/eu_demo_2015_reactor.py
```diff
@@ -0,0 +1,421 @@
...
+    """Creates geometry of a simplified EU DEMO model based on the published
+    diagram in Figure 2 of Definition of the basic DEMO tokamak geometry based
+    on systems code studies. Published in Fusion Engineering and Design
+    http://dx.doi.org/10.1016/j.fusengdes.2015.06.097 . Coordinates extracted
+    from the figure are not exact and therefore this model does not perfectly
+    represent the reactor.
+
+    Arguments:
+        rotation_angle (float): the angle of the sector that is desired.
+            Defaults to 360.0.
```
shimwell:
```suggestion
        number_of_tf_coils: the number of tf coils to include in the model
```

paramak/parametric_reactors/eu_demo_2015_reactor.py
```diff
@@ -0,0 +1,421 @@
...
+class EuDemoFrom2015PaperDiagram(paramak.Reactor):
+    """Creates geometry of a simplified EU DEMO model based on the published
+    diagram in Figure 2 of Definition of the basic DEMO tokamak geometry based
+    on systems code studies. Published in Fusion Engineering and Design
+    http://dx.doi.org/10.1016/j.fusengdes.2015.06.097 . Coordinates extracted
+    from the figure are not exact and therefore this model does not perfectly
+    represent the reactor.
+
+    Arguments:
+        rotation_angle (float): the angle of the sector that is desired.
```
shimwell:
```suggestion
        rotation_angle: the angle of the sector that is desired.
```

paramak/parametric_reactors/eu_demo_2015_reactor.py
```diff
@@ -0,0 +1,421 @@
...
+            Defaults to 360.0.
+    """
+
+    def __init__(
+        self,
+        rotation_angle=360.0,
+        number_of_tf_coils=11,
```
shimwell:
```suggestion
        rotation_angle: float = 360.,
        number_of_tf_coils: int = 16,
```

paramak/parametric_reactors/eu_demo_2015_reactor.py
```diff
@@ -0,0 +1,421 @@
...
+                (588.9188915604805, -442.35157799696844),
+                (578.1466100551762, -383.4976698929381),
+                (577.6991331872189, -343.8423522802541),
+                (576.2157294079991, -254.96634344146696),
+                (576.2157294079991, -201.97055742152952),
+                (576.2157294079991, -148.97477140159208),
+                (576.2157294079991, -95.97898538165487),
+            ],
+            rotation_angle=self.rotation_angle,
+            # avoid overlap between VV and blanket divertor
```
shimwell, line 222:
```suggestion
```

paramak/parametric_reactors/eu_demo_2015_reactor.py
```diff
@@ -0,0 +1,421 @@
...
+            rotation_angle=self.rotation_angle,
+            stp_filename='vacvessel.stp',
+            stl_filename='vacvessel.stl',
+        )
+
+        return [divertor, blanket, vac_vessel, vac_vessel_inner]
+
+    def create_plasma(self) -> list:
+        """Creates a 3d solids for the plasma.
+        Returns:
```
shimwell:
```suggestion
        """Creates a 3d solids for the plasma.

        Returns:
```

paramak/parametric_reactors/eu_demo_2015_reactor.py
RemDelaporteMathurin: Ah yes forgot that one. Don't we want to keep the Defaults to 360.0 though ?

paramak/parametric_reactors/eu_demo_2015_reactor.py
RemDelaporteMathurin: This line was to expplain the union of blanket and divertor. If you want to get it of it it's fine with me

shimwell: ah ok ignore this suggestion then

paramak/parametric_reactors/eu_demo_2015_reactor.py
shimwell: The default is in the argument and appears with type hinting. I think we only need to add "Defaults to X" when there are specific consequences. Otherwise we are just duplicating text. There are a few cases where we have Defaults to None which require further explanation but in this case I think in this case the default in the args is fine.

Happy to go with whatever you like but that is my train of thought

shimwell:
I think we just need a FreeCAD picture with a similar color scheme to the others, but I can do that when I build it over here

shimwell:
We should also update the grid of 6 reactors that we have but could do with one more reactor to make a 3 by 3 grid

Screenshot from 2021-03-06 23-24-24

RemDelaporteMathurin:
> We should also update the grid of 6 reactors that we have but could do with one more reactor to make a 3 by 3 grid
> 

Real nice! Are you talking about the Figure 4 of the paper ? If so are you thinking about adding SPARC, DEMO and another one ?

We could try ITER...

shimwell (approved):

Issue #737, opened 2021-02-19
Automatically replacing the Circles with Spines to allow sector models


shimwell:
When simulating a 360 model the conversion of STEP into h5m with Trelis appears to be fine. :+1: 

However when creating a sector model it appears that Trelis (version 17.1.0) loads up the stp files and sections get broken in the process. I've looked at all the sections that get broken and noticed they are all relating to circle edges.

With some advice from Andy I now think that OCC circles don't always convert to ACIS circles reliably and it appears that I need a solution for removing circles from the geometry.

This is a shame as circles define the shape very well. However with the recently introduced facing tools this shouldn't be too much of a challenge to internally convert circular edges with splines and get a very close match.

example of broken geometry
![Screenshot from 2021-02-17 19-47-35](https://user-images.githubusercontent.com/8583900/108504590-080edd80-72ae-11eb-9e1b-7aa718a193b2.png)

shimwell:
I have been working on this method of replacing circle edges with splines by searching the Shape.points for circles and then creating the circle edges and then faceting them with splines and replacing them.

A working progress is on this branch

https://github.com/ukaea/paramak/tree/converting_circle_connections_to_splines

shimwell:
This is how the method works at he moment

```
import paramak

rotated_mixed = paramak.RotateMixedShape(
    rotation_angle=180,
    points=[
        (100, 0, "straight"),
        (200, 0, "circle"),
        (250, 50, "circle"),
        (200, 100, "straight"),
        (150, 100, "straight"),
        (140, 75, "circle"),
        (110, 45, "circle"),
    ]
)
rotated_mixed.convert_circle_edges_to_splines(tolerance=1)
```

shimwell:
#748 added this capability, but it didn't fix the problem

PR #748, opened 2021-02-26
added method to convert circle edges to spline edges

shimwell:
## Proposed changes

as described in issue #737 circle edges can cause problems when importing the stp files into trelis.

This PR provides the user with a method of replacing the circles with spines of specified tolerance.

TODO

- [x] adding tests
- [x] adding docstrings
- [x] perhaps move from shape.py to mixed shapes or utils

## Types of changes

- [ ] Code refactoring
- [x] New tests

## Checklist

- [x] Pep8 applied

### DIFF
--- a/paramak/utils.py
+++ b/paramak/utils.py
@@ -659,7 +659,7 @@ def convert_single_circle_to_spline(
         p1: Tuple[float, float],
         p2: Tuple[float, float],
         tolerance: Optional[float] = 0.1
-        ) -> List[Tuple[float, float, str]]:
+) -> List[Tuple[float, float, str]]:
     """Converts three points on the edge of a circle into a series of points
     on the edge of the circle. This is done by creating a circle edge from the
     the points provided (p0, p1, p2), facets the circle with the provided
--- a/paramak/utils.py
+++ b/paramak/utils.py
@@ -686,18 +686,15 @@ def convert_circle_to_spline(
         edges=new_edge,
         view_plane='XZ'
     )
-    points_with_connections = []
-    for point in points:
-        points_with_connections.append((point[0], point[1]))
 
-    return points_with_connections
+    return points
 
 
 class FaceAreaSelector(cq.Selector):
     """A custom CadQuery selector the selects faces based on their area with a
     tolerance. The following useage example will fillet the faces of an extrude
-    shape with an area of 0.5. paramak.ExtrudeStraightShape(points=[(1,1),(2,1),
-    (2,2)], distance=5).solid.faces(FaceAreaSelector(0.5)).fillet(0.1)
+    shape with an area of 0.5. paramak.ExtrudeStraightShape(points=[(1,1),
+    (2,1), (2,2)], distance=5).solid.faces(FaceAreaSelector(0.5)).fillet(0.1)
 
     Args:
         area (float): The area of the surface to select.
--- a/paramak/utils.py
+++ b/paramak/utils.py
@@ -663,8 +663,7 @@ def convert_circle_to_spline(
     """Converts three points on the edge of a circle into a series of points
     on the edge of the circle. This is done by creating a circle edge from the
     the points provided (p0, p1, p2), facets the circle with the provided
-    tolerance to extracts the points on the faceted edge and returns them with
-    spline connections.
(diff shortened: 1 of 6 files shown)


shimwell:
Example usage case. This will convert circles to splines and add extra points so the splines accurately reflect the circle shape. Default tolerance is 0.1 but can be user specified

```
rotated_mixed = paramak.RotateMixedShape(
    rotation_angle=180,
    points=[
        (100, 0, "straight"),
        (200, 0, "circle"),
        (250, 50, "circle"),
        (200, 100, "straight"),
        (150, 100, "straight"),
        (140, 75, "circle"),
        (110, 45, "circle"),
    ]
)
rotated_mixed.convert_all_circle_points_to_splines(tolerance=0.01)
```

shimwell:
tests, doc strings and reorganisation is finished. All the checks and tests are passing.

I think this is ready to be merged. If there are no objects i can merge this in on Tuesday at midday

RemDelaporteMathurin (commented):
@Shimwell This looks tidy. Is there a way to test that transforming circles to edges doesn't modify the shape too much ? For instance by creating a circle shape, converting it to splines, and check that the difference in volumes is low ?

RemDelaporteMathurin:
> @Shimwell This looks tidy. Is there a way to test that transforming circles to edges doesn't modify the shape too much ? For instance by creating a circle shape, converting it to splines, and check that the difference in volumes is low ?
> 
> Great idea @RemDelaporteMathurin I was silly and testing for the code coverage more than actually designing good tests. I shall add the suggested test, thanks for the idea

@Shimwell seems like you've been editing my comment rather than replying to it therefore I hadn't seen it ;-)

shimwell:
Great idea @RemDelaporteMathurin I was silly and testing for the code coverage more than actually designing good tests. I shall add the suggested test, thanks for the idea

Screen shot shows the circle edge from the test shape overlaid with the spine replacement edge. This is with the default tolerance (0.1) and there is very little difference (less than 0.000001 rel difference in volume)
Screenshot from 2021-02-27 13-25-44


@eepeterson @jbae11 looks like circles can easily be replaced with splines and the volumes kept to approximately the same value. Hoping this fixes #737

But perhaps more interestingly we could use similar a edge replacement method to find splines and replaced them with a series of straight lines and make spline free geometry for easier conversion to CSG. Perhaps something for the future

tests/test_utils.py
```diff
@@ -12,6 +12,33 @@
...
+        new_points = paramak.utils.convert_circle_to_spline(
+            p0=(200, 0),
+            p1=(250, 50),
+            p2=(200, 100),
+            tolerance=0.2
+        )
+
+        connections = [connection[2] for connection in new_points]
+
+        assert len(set(connections)) == 1
```
shimwell:
```suggestion
        assert new_points[0][0] == 200
        assert new_points[0][1] == 0
        assert new_points[-1][0] == 200
        assert new_points[-1][1] == 100
        assert len(set(connections)) == 1
```

RemDelaporteMathurin (commented):
I wonder if the tests should live in test_parametric_shapes since they seem to apply to RotateMixedShape and ExtrudeMixedShape and not only to the super class Shape ?

paramak/utils.py
```diff
@@ -654,6 +654,45 @@ def export_wire_to_html(
...
+
+    new_edge = paramak.utils._transform_curve(edge, tolerance=tolerance)
+
+    points = paramak.utils.extract_points_from_edges(
+        edges=new_edge,
+        view_plane='XZ'
+    )
+    points_with_connections = []
+    for point in points:
+        points_with_connections.append((point[0], point[1]))
```
RemDelaporteMathurin: what's the difference between `points` and `points_with_connections` ?

RemDelaporteMathurin (approved):

paramak/utils.py
shimwell: ah  yes, sorry I think they were different early on in this PR but now they are the same. I can remove this.

shimwell:
> I wonder if the tests should live in test_parametric_shapes since they seem to apply to RotateMixedShape and ExtrudeMixedShape and not only to the super class Shape ?

I've moved the tests over now and I agree it is better than before

Issue #785, opened 2021-03-25
Incorrect ToroidalFieldCoilTripleArc R1 variable


billingsley-john:
Documentation suggests R1 is the distance to the rear/outer of the inboard leg of the TF coil. However, upon construction, this parameter is the distance to the front/inner of the inboard leg. Not sure whether the documentation image or construction is incorrect.

`
shape = p.ToroidalFieldCoilTripleArc(
    R1=40,
    h=200,
    radii=(70, 100),
    coverages=(60, 60),
    thickness=20,
    distance=40,
    number_of_coils=1
)
`

Notice x-value for rear/outer points of inboard leg = 20, when should be 40 (according to documentation)

![Screenshot from 2021-03-25 12-31-08](https://user-images.githubusercontent.com/56687624/112473247-041b2300-8d66-11eb-9c42-b26dc3b35938.png)

![Screenshot from 2021-03-25 12-43-35](https://user-images.githubusercontent.com/56687624/112474600-c15a4a80-8d67-11eb-8ee7-7573928ddc1e.png)

shimwell:
Nice find John. I guess it is useful to have the variable set to that inner distance, perhaps we should changed the code so that the thickness is added to R1 when finding the points.

PR #786, opened 2021-03-25
updated R1 variable ToroidalFieldCoilTripleArc

billingsley-john:
## Proposed changes

PR which updates the ToroidalFieldCoilTripleArc class so that the R1 variable is used correctly, as described in the documentation.

## Types of changes

- [ ] Code refactoring
- [ ] New tests

## Checklist

- [x] Pep8 applied

## Further comments

### DIFF
--- a/paramak/parametric_components/toroidal_field_coil_triple_arc.py
+++ b/paramak/parametric_components/toroidal_field_coil_triple_arc.py
@@ -115,12 +115,12 @@ class ToroidalFieldCoilTripleArc(ExtrudeMixedShape):
 
         # create inner coordinates
         R_inner, Z_inner = self._compute_curve(
-            self.R1, self.h * 0.5, radii=(small_radius, mid_radius),
+            self.R1 + thickness, self.h * 0.5, radii=(small_radius, mid_radius),
             coverages=(small_coverage, mid_coverage))
 
         # create outer coordinates
         R_outer, Z_outer = self._compute_curve(
-            self.R1 - thickness, self.h * 0.5,
+            self.R1, self.h * 0.5,
             radii=(small_radius + thickness, mid_radius + thickness),
             coverages=(small_coverage, mid_coverage))
         R_outer, Z_outer = np.flip(R_outer), np.flip(Z_outer)


billingsley-john:
Resolves #785

billingsley-john:
ToroidalFieldCoilPrincetonD has a similar issue; however, is documented correctly. How do you feel @Shimwell @RemDelaporteMathurin about this? I think both R1 values in both classes should correspond to the same distance, it is a question of whether it's the inner or outer distance?
One thing to note is that TFCoilPrincetonD also has R2, which is an 'inner' radius, perhaps, R1 should do the same, which would suggest updating the image for TfCoilTripleArc rather than the class.

shimwell:
I have a slight preference for inner distance as I think it makes the radial build easier and allows us to keep the same diagram

RemDelaporteMathurin:
> ToroidalFieldCoilPrincetonD has a similar issue; however, is documented correctly. How do you feel @Shimwell @RemDelaporteMathurin about this? I think both R1 values in both classes should correspond to the same distance, it is a question of whether it's the inner or outer distance?

I agree with @Shimwell , inner is best.

> One thing to note is that TFCoilPrincetonD also has R2, which is an 'inner' radius, perhaps, R1 should do the same, which would suggest updating the image for TfCoilTripleArc rather than the class.

I do agree that R1 should be a inner radius as it makes it easier - I think - to build with it.

billingsley-john:
> > ToroidalFieldCoilPrincetonD has a similar issue; however, is documented correctly. How do you feel @Shimwell @RemDelaporteMathurin about this? I think both R1 values in both classes should correspond to the same distance, it is a question of whether it's the inner or outer distance?
> 
> I agree with @Shimwell , inner is best.
> 
> > One thing to note is that TFCoilPrincetonD also has R2, which is an 'inner' radius, perhaps, R1 should do the same, which would suggest updating the image for TfCoilTripleArc rather than the class.
> 
> I do agree that R1 should be a inner radius as it makes it easier - I think - to build with it.

I agree with you both - as in R1 as outlined in the documentation.

shimwell (approved):
I agree with you all as well :-)

PR #789, opened 2021-03-25
updated R1 variable ToroidalFieldCoilPrincetonD

billingsley-john:
## Proposed changes

PR updating the R1 variable of the TFCoilPrincetonD to correspond to the radial distance from x=0 to the first layer of the inner leg of the tf coil.

## Types of changes

- [x] Code refactoring
- [ ] New tests

## Checklist

- [ ] Pep8 applied

## Further comments

### DIFF
--- a/paramak/parametric_components/tf_coil_casing.py
+++ b/paramak/parametric_components/tf_coil_casing.py
@@ -31,7 +31,7 @@ class TFCoilCasing(ExtrudeMixedShape):
         material_tag: Optional[str] = "TF_coil_casing_mat",
         **kwargs
     ) -> None:
-    
+
         self.magnet = magnet
 
         super().__init__(
--- a/paramak/parametric_components/tf_coil_casing.py
+++ b/paramak/parametric_components/tf_coil_casing.py
@@ -11,9 +11,9 @@ class TFCoilCasing(ExtrudeMixedShape):
 
     Args:
         magnet: TF coil shape
-        inner_offset: radial distance between inner coil surface and inner 
+        inner_offset: radial distance between inner coil surface and inner
             casing surface (cm)
-        outer_offset: radial distance between outer coil surface and outer 
+        outer_offset: radial distance between outer coil surface and outer
             casing surface (cm)
         vertical_section_offset: radial distance between outer coil surface and
             outer vertical section surface (cm)
@@ -21,9 +21,9 @@ class TFCoilCasing(ExtrudeMixedShape):
     """
 
     def __init__(
-        self, 
-        magnet, 
-        inner_offset: float, 
+        self,
+        magnet,
+        inner_offset: float,
         outer_offset: float,
         vertical_section_offset: float,
         stp_filename: Optional[str] = "TFCoilCasing.stp",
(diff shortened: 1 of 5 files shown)


billingsley-john:
This has been achieved by simply adding + self.thickness to the arguments of self._compute_inner_points(); however, this can cause confusion as R1 within that method is no longer equal to self.R1 - perhaps this method should be updated to use a different variable name.

shimwell:
I wonder if it is worth making the ITERTokamak and  EuDemoFrom2015PaperDiagram to check their TF coils are and casing are still working

billingsley-john:
> I wonder if it is worth making the ITERTokamak and EuDemoFrom2015PaperDiagram to check their TF coils are and casing are still working

Indeed, ITERTokamak will need updating

shimwell (approved):

billingsley-john:
As vertical_section_offset can now be smaller than outer_offset, the method of construction where the outer coil points are translated by outer_offset and unioned with the inner leg means excess at the back of the casing can exist (see image). Previously, when the inner radius of the inner leg was determined by outer_offset + vertical_section_offset, this would not happen as the translated outer points would always be within the inner leg.

Screenshot from 2021-03-26 15-59-21

This can be solved by cutting away the excess coil.

billingsley-john:
New functionality

`
casing = paramak.TFCoilCasing(magnet=magnet, inner_offset=10, outer_offset=30, vertical_section_offset=20, distance=40)
`

Screenshot from 2021-03-26 16-15-29

billingsley-john:
Not the most elegant solution so any suggestions welcome!

billingsley-john:
TODO:
Add tests

billingsley-john:
TFCoilPrincetonD

shimwell:
Looks like there is just one tiny code inspector comment

paramak/parametric_components/tf_coil_casing.py
```diff
@@ -100,5 +141,13 @@ def create_solid(self):
             self.azimuth_placement_angle
         solid = union_solid(solid, self.leg_shape)
         solid = cut_solid(solid, self.magnet)
+
+        inner_bore_cutting_shape = ExtrudeStraightShape(
```
RemDelaporteMathurin, line 148: @billingsley-john perhaps some comments could help understand what you did here (nice find btw)

RemDelaporteMathurin: maybe even refer to the issue

RemDelaporteMathurin (changes requested):

RemDelaporteMathurin:
Nice find @billingsley-john. If I understand well, you have an additionnal cut operation to get rid of any potential excess right ?

paramak/parametric_components/tf_coil_casing.py
billingsley-john: Sounds great, I shall add some!

billingsley-john:
> Nice find @billingsley-john. If I understand well, you have an additionnal cut operation to get rid of any potential excess right ?

Yes, there is sometimes excess magnet depending on the outer_offset and vertical_section_offset values. This cuts away this excess. Before merging, we should test the component is still created successfully when there is no excess. I'm not entirely sure of the behaviour of CadQuery when a cut is attempted but the solids are not overlapping. Perhaps this has to be guarded against (i.e. only perform a cut if there is excess)

RemDelaporteMathurin (approved):

billingsley-john:
Now uses coil workplane to create casing in correct workplane

